PR #12, opened 2013-10-25
New plugin: (DEIND) JPverbLite (similar sound, fewer features but 50% CPU usage of the full version)

totalgee:
I really like the new JPverb (in DEIND UGens), but wanted a lighter-weight version, for cases where I don't need all its functionality.  For this version, I just took JPverb's Faust code and removed elements (in particular, the delay-line chorusing feature and the low/mid/high frequency reverb time scaling).  I also reduced the number of reflections somewhat.  This "light" version uses about 50% of the CPU of the original version, but still sounds very similar to the original (at least, when modulation depth is 0 and the frequency-band reverb time multipliers are all set to 1).  This makes it good for cases when you want a nice-sounding stereo reverb without too much processing power.

You'd need to combine two GVerb plugins to build a "stereo input" reverb, and a single instance of this new JPverbLite UGen (which handles stereo input) uses slightly less CPU than two GVerb plugins (and it sounds much better, IMHO).

Thanks to Julian Parker, Till Bovermann and Stefan Kersten for their work on the original plugin! (this is just a stripped-down version, nothing new here)  If they don't mind, it would be handy to have this available as part of the DEIND UGens collection.


telephon:
I just asked, as the patch removes custom settings which had been carefully tuned,it shouldn't be merged. But it can remain as a branch of course.

LFSaw:
thanks for this, @telephon .
as said, please feel free to use JPVerb (and all the other UGens) as a basis for your own implementations and post them in your own packages. However, we prefer to keep the DEIND UGen package as it is.

PR #30, opened 2015-03-01
ATK: Generalize FoaEncode and FoaDecode behavior

backseatviolist:
Currently FoaEncode/FoaDecode only accept encoders and decoders that are strict members of four ATK classes. I ran into this problem when trying to extend them, so this change allows them to accept subclasses.

### DIFF
--- a/source/ATK/sc/Classes/ATK.sc
+++ b/source/ATK/sc/Classes/ATK.sc
@@ -459,9 +459,9 @@ FoaUGen {
 FoaDecode : FoaUGen {
 	*ar { arg in, decoder, mul = 1, add = 0;
 		in = this.checkChans(in);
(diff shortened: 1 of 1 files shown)


joslloand:
Ok, I see that Nathan's request is about making the ATK easier to work with Ctk. Nathan's comment is on this thread:

http://new-supercollider-mailing-lists-forums-use-these.2681727.n2.nabble.com/ATK-seemingly-inconsistency-td7618884.html#a7618892

Here's the relevant comment:

> There's a way to get ATK's kernel encoders and decoders working in Ctk without booting another ? server: https://github.com/backseatviolist/ctk-atk

We may wish to merge the above Ctk functionality directly into the ATK.

PR #31, opened 2015-03-04
59837f4 need C++11 to build std::trunc from 7a82fdb

andersvi:


bagong:
Any reasons not to commit this?

telephon:
I can't see any

bagong:
Then why don't you press the button? :)

telephon:
just because i am potentially blind

bagong:
Hahaha, I don't think so, and there is always revert ;) Saved me typing -DCPP11=ON about 10 times today ;) (and trying to remember if it's c++11 or cpp11... Thanks!

PR #34, opened 2015-03-26
Add cmake arguments to .travis.yml that are tested on Mac and Ubuntu

bagong:
I added cmake arguments that work on a Mac 10.10 and Ubuntu 14.04

Problems:
- BhobUgens currently break the build (see issue)
- Linux install insists on an linux file system logic, so it will install to
    `share/SuperCollider/Extensions/SC3...`
    starting from the CMAKE_PREFIX_PATH  
    I didn't understand how this is handled in the SC travis script, but it's not used here anyways atm

Note: 
- on Mac if you specify a cmake prefix path you must add IN_PLACE_BUILD=OFF
- CPP11=ON is required now
- I think the CMakeLists.txt needs an overhaul, but I didn't change anything, just added cmake-arguments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ cache:
 before_install:
  - ifmac () { if [[ $TRAVIS_OS_NAME == osx ]]; then eval $@; fi; }
  - iflin () { if [[ $TRAVIS_OS_NAME == linux ]]; then eval $@; fi; }
+ - ifmac brew update
  - ifmac brew install cmake || true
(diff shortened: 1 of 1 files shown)


bagong:
Yes the build breaks in BhobUGens now. That's the same I get locally. So it can be expected that this works, once the BhobUGens issue is adjusted. All other plugins build for me locally.

gusano:
> Yes the build breaks in BhobUGens now. That's the same I get locally

On which platform/OS?

bagong:
OSX

bagong:
Allready done in a more sophisticated way ;)

PR #36, opened 2015-04-26
atk: avoid branches in inner loop and memory hazards

timblechmann:
branches and memory are MUCH slower than doing the actual computation.
this patch speeds roughly doubles the performance of atk panners

### DIFF
--- a/source/ATK/AtkUGens.cpp
+++ b/source/ATK/AtkUGens.cpp
@@ -265,27 +265,25 @@ extern "C"
 }
 
 inline float calcmatrixval(float coef, float curval){
-    float val;
-    if(coef == 0.){
-	val = 0.;
-    } else {
-	if(coef == 1.){
-	    val = curval;
-	} else {
-	    val = coef * curval;
-	}
-    }
-    return val;
+    return coef * curval;
 }
 
 // can perhaps optimize a bit here ... check for 0s???
 #define CALC_MATRIX \
 float curvals[4] = {Win[i], Xin[i], Yin[i], Zin[i]}; \
 for(int j = 0; j < 4; j++){ \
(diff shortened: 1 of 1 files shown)


joshpar:
Hi Tim,
The patch looks fine, but it will drastically increase the number of multiplies needed to calculate most of the matrices. Have you tested it along with the transforms as well?
Otherwise, looks fine to me.
Best
Josh

/*
Josh Parmenter
www.realizedsound.net/josh
*/

timblechmann:
it might drastically increase the number of multiplies, but also drastically reduces the number of branches, which have a much bigger impact on performance.

joshpar:
Go for it

/*
Josh Parmenter
www.realizedsound.net/josh
*/

telephon:
>Go for it

Josh, can you merge it?

bagong:
I breaks the Mac build at the same place where it happened before Scott C added a workaround.

bagong:
Ah, while the build broke, that was obviously due to the current clang version when I tested the build, which must have differed from Scott C's the one specified in Scott's workaround (6.0.0.600005) for the vectorization problem. So there is no connection to this PR.
The good news: the clang bug is gone in v. 6.1!
So no build-prob because of this PR!! Sorry for the misinterpretation.

bagong:
>> Go for it

> Josh, can you merge it?

@joshpar, I don't think anybody will merge it except you ;)

joslloand:
Hi All,

I can merge... I'd be ready to merge the request, but I'm seeing:

>Merge with caution!
>Failed — The Travis CI build failed

As a result, I'm hesitant to push go. Any advice here?

bagong:
The build error is not related to the PR. It was because of the compiler version used by travis at the time the PR was submitted. I just retriggered the build and it works fine.

joslloand:
Merged...

PR #40, opened 2015-10-15
Libmodify

sonoro1234:
implement multitap delay in dwglib and kl-junctions


telephon:
Thank you for these - do you have some documentation, too? As a stub, even a normal ````.scd```` file is often helpful even.

Issue #44, opened 2015-11-25
outdated stk sources included with tarball


hzulla:
The [stk source copy](https://github.com/supercollider/sc3-plugins/tree/master/source/StkUGens/stk-4.4.2) that comes with the sc3 plugins is outdated. The tarball contains 4.4.2, last [stk release is 4.5.0](https://github.com/thestk/stk/) more than a year ago.

danstowell:
thanks for pointing this out. does it cause a problem?

hzulla:
It's more a janitorial issue, I guess.

I'm currently trying to [package Sonic Pi for Debian/Ubuntu](https://launchpad.net/~hzulla/+archive/ubuntu/sonic-pi), which needs the sc3 plugins, so those got packaged as well.

Debian packaging rules tell packagers to [avoid embedded source code copies](https://www.debian.org/doc/debian-policy/ch-source.html#s-embeddedfiles) but use existing Debian packages whenever possible.

Debian comes with stk-4.4.4. I tried to package the sc3 plugins with the Debian library, but they have a [small error](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=805549) with their current package. Once that one is resolved, it is indeed possible to build, link and package the sc3 plugins against the Debian library.

The Debian multimedia packagers told me on irc that they may switch to stk-4.5.0 soon, so this issue is just to notify you about it.

hzulla:
stk-4.5.0 changed some filenames, so some `include` lines of the Stk plugin may need to be updated for it.

sonoro1234:
In issues #33 I anounced a new Stk wrapping as the old one is not working.
Nobody has aswered

crucialfelix:
I can understand using the debian stk if this was a debian-only package.

But since sc plugins are crossplatform I think it only causes more
maintenance problems to have compilation different for one platform.
In this case I think included source code is better.

my 2c

danstowell:
@crucialfelix we already handle it for debian in supercollider's main codebase: the debian packaging scripts patch SC to use system boost rather than the bundled one. this happens all in "downstream" and the SC developers never need to know. the same thing would happen here, we don't need to worry about it in our upstream.

[...]

PR #46, opened 2015-11-26
Update STK from 4.4.2 to 4.4.4

danstowell:
See issue #44. I've gone for version 4.4.4 (not 4.5.0) since that's a minor API change and also matches current debian.

I've tested that it builds OK but not that it works, since I'm not an STK user.


danstowell:
Merging this since it's a minor versionbump and the "new" Stk wrapper is also there too

telephon:
on cmake, I get the following error:

```
CMake Error at source/StkInst/CMakeLists.txt:22 (list):
  list sub-command REMOVE_ITEM requires two or more arguments.
```

danstowell:
@telephon that's not an issue with this PR. It's caused by #47, simply because you need to run `git submodule init` since #47 adds a new submodule. (NB: I believe the obscure error message that you quote is already fixed, replaced with an informative one...) Please add a comment on #47 if the problem persists for you.

telephon:
ah good - this fixed it.

PR #47, opened 2015-11-27
StkInst into sc3plugins

sonoro1234:
This builds StkInst but breaks StkUGens compilation because there is only one CMAKE file and include_directories from both projects get mixed


sonoro1234:
I am getting this error even with master branch:
```
...\BerlachUGens.cpp:296:25: error: 'M_PI' was not declared in this scope
    double w0 = freq * 2*M_PI / SAMPLERATE;
```

sonoro1234:
@danstowell  Works now in travis but to allow c11++ BerlachUGens should define M_PI

danstowell:
I'd suggest that should be a separate PR. I don't know why you need that change - I don't need it. (I just tested by rebuilding from scratch.)

sonoro1234:
@danstowell It is not needed by this PR (already finished and waiting for merge)

But it is needed with c++11 mingw 4.8.2:
without c++11 there is a complain about std::trunc but with c++11 there is a complain about M_PI not defined.
What is your compiler and c++11 setting?

source/StkInst/README.md
```diff
@@ -0,0 +1,2 @@
+# StkInst
+Stk wraping for SC
```
danstowell: is this file needed? also, spelling should be "wrapping"

source/StkInst/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
+cmake_minimum_required (VERSION 2.8)
+set(PROYECTO "StkInst")
```
danstowell: might be best to stick with english variable names, for future maintainance?

source/StkInst/README.md
sonoro1234: I am going to delete it!!

source/StkInst/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+
+include_directories(${SC_PATH}/include/plugin_interface)
+include_directories(${SC_PATH}/include/common)
+include_directories(${SC_PATH}/common)
+include_directories(${SC_PATH}/external_libraries/libsndfile/)
+
+
+set(CMAKE_SHARED_MODULE_PREFIX "")
+if(APPLE OR WIN32)
+set(CMAKE_SHARED_MODULE_SUFFIX ".scx")
```
danstowell: the indentation in this file is inconsistent. could I request indenting inside the if statements please?

source/StkInst/StkInst.cpp
```diff
@@ -0,0 +1,421 @@
...
+
+// The order of the following list is important.  The location of a particular
+// instrument in the list should correspond to that instrument's ProgramChange
+// number (i.e. Clarinet = ProgramChange 0).
+char insts[NUM_INSTS][10] = { "Clarinet", "BlowHole", "Saxofony", "Flute", "Brass",
+                              "BlowBotl", "Bowed", "Plucked", "StifKarp", "Sitar", "Mandolin",
+                              "Rhodey", "Wurley", "TubeBell", "HevyMetl", "PercFlut",
+                              "BeeThree", "FMVoices", "VoicForm", "Moog", "Simple", "Drummer",
+                              "BandedWG", "Shakers", "ModalBar", "Mesh2D", "Resonate", "Whistle" };
+#define GETMEM(Inst) RTAlloc(gWorld, sizeof(Inst))
```
danstowell: I would suggest an inline function is better than a #define here, that's a more modern and debuggable approach. (I know SC's codebase has a lot of #define in it...)

source/StkInst/StkInst.cpp
```diff
@@ -0,0 +1,421 @@
...
+	unit->gate = false;
+	unit->ampat	= IN0(2);
+	unit->XD = IN0(3);
+	unit->YD = IN0(4);
+	unit->Xpos = IN0(5);
+	unit->Ypos = IN0(6);
+	unit->decay = IN0(7);
+    
+    void* mem = RTAlloc(unit->mWorld,sizeof(Mesh2D));
+    if (!mem) {
```
danstowell: (indentation goes a bit odd here)

source/StkInst/StkInst.cpp
sonoro1234: Would not be a function but a static global variable?

danstowell:
I've made some comments from reading the code. They're not demands, just suggestions, since sc3plugins is supposed to be more permissive than the main SC3 code. But see what you think. Then I'll do a basic test if the PR seems to work for me and if so I'll merge it.

source/StkInst/StkInst.cpp
sonoro1234: do you mean?
```
+	unit->ampat	= IN0(2);
```

danstowell:
One more request, if that's OK! If the stk submodule is not checked out, please could you detect that and show an error message? Otherwise it gives a cryptic error "list sub-command REMOVE_ITEM requires two or more arguments" and users will get confused.

sonoro1234:
How can this be done?

source/StkInst/StkInst.cpp
danstowell: I don't see why?

danstowell:
http://stackoverflow.com/a/13089406/1563927

source/StkInst/StkInst.cpp
sonoro1234: sorry thought you meant NUM_INSTS
going to change it!!

sonoro1234:
```
void* GETMEM(Instrmnt inst){
   return  RTAlloc(gWorld,sizeof(inst));
}
```
Would work!!??

danstowell: Yes but add "inline" to the start

sonoro1234:
```
C:\SupercolliderRepos\pluginfork\sc3-plugins\source\StkInst\StkInst.cpp:61:23: error: cannot declare parameter 'Inst' to be o
f abstract type 'stk::Instrmnt'
 void* GETMEM(Instrmnt Inst){
```
Seems not to work. Better leave it as a #define

danstowell:
This should probably be fixed please:
```
WARNING: SCDoc: In /usr/local/share/SuperCollider/Extensions/SC3plugins/StkInst/HelpSource/Classes/Stk.schelp
  Method *at has arg named 'preset', but doc has 'argument:: name'.
```

danstowell:
Thanks for contributing this Victor!

sonoro1234:
You are welcome!!

PR #48, opened 2015-11-29
BerlachUGens: add M_PI definition for c11++ compiling

sonoro1234:


timblechmann:
fwiw, https://stackoverflow.com/questions/1727881/how-to-use-the-pi-constant-in-c

sonoro1234:
@timblechmann  what would be your prefered solution?

timblechmann:
no strong opinion, but the preprocessor define before including math.h seems a bit cleaner

sonoro1234:
but this comment is confusing
```
Even after defining _USE_MATH_DEFINES if GCC complains that's because __STRICT_ANSI__ is defined (perhaps you passed -pedantic or -std=c++11) which disallows M_PI to be defined, hence undefine it with -D__STRICT_ANSI__. When defining it yourself, since it's C++, instead of a macro you should constexpr auto M_PI = 3.14159265358979323846;. – legends2k Jan 23 '14 at 10:44 
```

PR #50, opened 2015-12-03
StkInst: cmake warning for submodule update

sonoro1234:
plus a minor change for not compiling one not needed file


sonoro1234:
@danstowell hope this works

PR #53, opened 2015-12-13
Add cmake build option SYSTEM_STK

danstowell:
Based on patch by @hzulla:
https://github.com/deb-pkg-sonic-pi/pkg-supercollider-sc3-plugins/blob/master/debian/patches/02-debian-libstk.patch


hzulla:
Works nicely. Thanks. /cc @petterreinholdtsen

hzulla:
This is cool, thank you. It's great when upstream accepts a patch even before one submits it. :-)

PR #56, opened 2015-12-17
STK as a git submodule in libs/stk

hzulla:
This removes the embedded copy of STK-4.4.4 from StkUGens and the git submodule from StkInst.

Instead, this uses a git submodule for both, located in libs/stk (similar to nova-simd).

This builds for both with and without SYSTEM_STK.

Needs testing. @sonoro1234, could you please review this?


hzulla:
This also detects wether the system's STK version is using `SKINI.msg` or `SKINImsg.h`, the latter name has been used after the release of STK 4.5.0 in the current development version.

sonoro1234:
For StkInst seems ok to me.
I dont know why is StkUGens using SKINI

hzulla:
StkAll.cpp is using defines from `SKINImsg.h`, e.g.:
* [__SK_BowPressure_](https://github.com/supercollider/sc3-plugins/blob/89baee709b255e12ca8ba11d489e481762500286/source/StkUGens/StkAll.cpp#L647)
* [__SK_BowPosition_](https://github.com/supercollider/sc3-plugins/blob/89baee709b255e12ca8ba11d489e481762500286/source/StkUGens/StkAll.cpp#L651)
* [__SK_ModFrequency_](https://github.com/supercollider/sc3-plugins/blob/89baee709b255e12ca8ba11d489e481762500286/source/StkUGens/StkAll.cpp#L655)
* [__SK_ModWheel_](https://github.com/supercollider/sc3-plugins/blob/89baee709b255e12ca8ba11d489e481762500286/source/StkUGens/StkAll.cpp#L659)
* [__SK_AfterTouch_Cont_](https://github.com/supercollider/sc3-plugins/blob/89baee709b255e12ca8ba11d489e481762500286/source/StkUGens/StkAll.cpp#L663)

etc.

hzulla:
P.S.: The motivation for this patch is
a) cleaning up the source tree to resemble it the [supercollider repo](https://github.com/supercollider/supercollider) where all external libraries are also put into one shared directory.
b) making it easier to handle the embedded source copy when packaging the sc3 plugins for Debian.

hzulla:
Thanks.

PR #57, opened 2015-12-17
Move TLSF to libs/TLSF-2.4.6

hzulla:
This moves the TLSF library from the NovaDiskIO subdirectory to libs/

(There is no functional change in this patch. Just a janitorial cleanup to put external libraries in one single place of the repository. If this ~~and #56~~ gets merged, I'd propose to rename libs/ to  external_libraries/ later, to make the repo layout more like the [supercollider repo](https://github.com/supercollider/supercollider))


hzulla:
Thanks.

Issue #59, opened 2016-01-05
Really weird, glitchy audio with JPverb and Greyhole


backseatviolist:
Run the following and leave it running for a while:

    (
    { JPverb.ar(Saw.ar(100) * EnvGen.kr(Env.perc(0.01, 1), Impulse.kr(0))) * 0.1 }.play;
    { SinOsc.ar(440) * EnvGen.kr(Env.perc(0.01, 1), Impulse.kr(1)) * 0.1!2 }.play;
    )

Watch the CPU. For me, it stays at a consistent 7-8% for a while. After about one minute, it spikes to 50% and then quickly drops to less than 1%, and the audio from the *other* synth sounds weird and crackly.

This happens consistently when audio is fed into a JPverb or a Greyhole followed by a long stretch of complete silence. After the silence, all other audio output from SuperCollider sounds crackly. If audio is played to the reverb unit again, the audio returns to normal after a few seconds.

Even weirder, the crackliness does not seem to show up in Stethoscope, FreqScope, or s.record!

jamshark70:
Dollars to donuts it's a denormal-number problem.

backseatviolist:
This same problem happened with another reverb ugen I was writing myself. They have in common that they were written in FAUST, so this problem may ultimately stem from FAUST's lack of protection from denormals by default. Throwing in some equivalent of SC's `zapgremlins` will probably fix it.

sonoro1234:
denormal handling is now handled in audio driver callback.
Is scsynth an old version?
  ----- Original Message -----

danstowell:
@sonoro1234 if a ugen creates denormals internally, and performs calculations with them, the ugen itself needs to zap them

sonoro1234:
sc_SetDenormalFlags(); is used inside audio driver callback but it means that "all number processing" in this thread make denormals go to zero

sonoro1234:
sc_SetDenormalFlags is defined in SC_World.cpp:165 but only for __SSE__ or __SSE2__ 
I dont know which architectures dont have that

danstowell:
@sonoro1234 ah of course, sorry.

backseatviolist:
@sonoro1234 I have experienced this error in both 3.6 and 3.7alpha1. Since this seems architecture-dependent, I am using 32-bit Ubuntu.

timblechmann:
[...]

PR #63, opened 2016-01-13
set flags for sse

sonoro1234:
They were not setted.
This was discovered in a denormal numbers issue #59
@timblechmann I copied flags from supercollider but I am not sure why -mfpmath=sse is disabled in apple if compiler is not clang
```
+    if(NOT APPLE AND NOT CMAKE_COMPILER_IS_CLANG)
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mfpmath=sse")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mfpmath=sse")
+    endif()
```


sonoro1234:
@timblechmann Ping?

timblechmann:
i don't think there are any 32bit x86 cpus without sse unit. we're speaking about roughly 20 years old intel  deskop cpus and some weird x86 cpus for embedded machines, which won't have sse

sonoro1234:
So, the lines above, copied from CMakeList.txt in supercollider, are ok or not?

LFSaw:
I'd say go for inclusion. We follow the supercollider build process as a guideline in sc-users. the question of why  ```-mfpmath=sse``` is disabled in apple if compiler is not ```clang``` might be forwarded to the supercollider development issue tracker. A possible change there should be mirrored in ```sc-plugins```.

timblechmann:
googling for "clang mfpmath" gives me as the first hit: https://stackoverflow.com/questions/19600238/how-do-i-pass-mfpmath-sse-to-llvm-using-clang

the first answer is:
> You do not need this flag with clang. It defaults to "-mfpmath=sse" as soon as your processor does ? support it.

before asking a human, it might sometimes be faster to ask a computer

LFSaw:
Thanks for looking into this, @timblechmann
. So we can safely merge?

LFSaw:
thanks :)

PR #67, opened 2016-01-18
cmake: on ARM platforms explicitly prejudge endianness. (Fixes #65)

### LINKED ISSUE #65: StkInst compile fail on ARM (BeagleBone Black), blocks build
If I try to build sc3-plugins on BeagleBone Black (platform is based on ARM Cortex A8), the cmake fails with:

```
...
-- Building plugins for SuperCollider version: 3.7alpha1
-- local_sources: 
-- Check if the system is big endian
-- Searching 16 bit integer
-- Looking for sys/types.h
-- Looking for sys/types.h - not found
-- Looking for stdint.h
-- Looking for stdint.h - not found
-- Looking for stddef.h
-- Looking for stddef.h - not found
-- Check size of unsigned short
-- Check size of unsigned short - failed
-- Check size of unsigned int
-- Check size of unsigned int - failed
-- Check size of unsigned long
-- Check size of unsigned long - failed
CMake Error at /usr/share/cmake-2.8/Modules/TestBigEndian.cmake:44 (message):
  no suitable type found
Call Stack (most recent call first):
  source/StkInst/CMakeLists.txt:50 (TEST_BIG_ENDIAN)


-- Configuring incomplete, errors occurred!

```

So, this is "caused" by the new StkInst STK wrapper by @sonoro1234. But the eventual cause seems to be that cmake, on this ARM platform, can't successfully complete the TEST_BIG_ENDIAN operation. I've checked the very latest (3.4.1) cmake source code and I don't think their TEST_BIG_ENDIAN has changed to address this, so I think at present we're stuck: the project simply can't build, because it chokes when configuring stk, and there's no option to deactivate it.

If I comment out the test (but since the ARM is little-endian I enable the flag that the test is supposed to decide about) then cmake configures happily.

Possible solutions: detect the arm platform ourselves; or add a flag to turn off building with stk if needed.

System info:
* Linux arm 3.8.13xenomai-bone41 #2 Wed Mar 26 11:41:31 GMT 2014 armv7l GNU/Linux
* cmake version 2.8.11.1

danstowell:


timblechmann:
maybe adapt the code to let the compiler do the heavy-lifting? http://sourceforge.net/p/predef/wiki/Endianness/

also boost.predef provides endianness detection macros (kind of the industry-standard): http://www.boost.org/doc/libs/1_60_0/libs/predef/doc/html/predef/reference/other_macros.html

danstowell:
@timblechmann thanks but the problem is not with the compiler or the source code in this case - it's simply a limitation in current cmake that I'm addressing here. Cmake has a function to detect endianness but the function barfs on arm. STK (external library, i.e. "not our code") requires a flag to be set to tell it the endianness, and there's no way I'm poking around in STK code to try and get rid of the need for that flag.

danstowell:
Ah SOLVED, see https://github.com/supercollider/sc3-plugins/issues/65#issuecomment-172651862

PR #72, opened 2016-02-21
Add DrumTrack from BBCut2

snappizz:
I'm the current maintainer of Nick Collins' BBCut2, and I am trying to get it packaged as a quark. It contains around 50 classes for sclang and three ugens: AnalyseEvents2, AutoTrack, and DrumTrack. Since quarks are client-only extensions, I need to get these out of the BBCut2 repository.

AnalyseEvents2 is in sc3-plugins and AutoTrack is in core (now known as BeatTrack), but for some reason DrumTrack was left behind. So DrumTrack needs a new home, and sc3-plugins seems like a reasonable place to put it.

Sorry for submitting such a large PR...

### DIFF
--- a/source/BBCut2UGens/DrumTrack.cpp
+++ b/source/BBCut2UGens/DrumTrack.cpp
@@ -1,0 +1,2076 @@
+/*
+	SuperCollider real time audio synthesis system
+ Copyright (c) 2002 James McCartney. All rights reserved.
+	http://www.audiosynth.com
+ 
+ This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


telephon:
I can't find `DrumTrack` in the current plugins. So this won't work.

danstowell:
@telephon I don't understand your comment. This PR is precisely proposing to *add* DrumTrack, no?

telephon:
Ah I didn't see the diff of the source file, as it was omitted, so I thought it was just the help and class file.

PR #77, opened 2016-04-23
Fix crash in OteySoundBoard and OteyPiano on Linux

snappizz:


sonoro1234:
great catch. Thanks Nathan.

PR #78, opened 2016-04-25
Windows: clarify install procedure for 32- and 64-bit release

bagong:


bagong:
Will adjust Readme to new situation and suggest two additions to general build configuration for Windows.

PR #80, opened 2016-04-29
Win msvc fixes

llloret:
Changes to make the sc3-plugins compile with MSVC 64 bits under Windows.

They should not break the other platforms, but I cannot test that.

To compile I am using the vanilla CMakeLists.txt, and passing -D options for the fftw library as explained in the README for Windows for SuperCollider.

Most plugins are compiling now, but there are still some that give errors when compiling. I tested with:
`{ BlitB3.ar(XLine.kr(10000, 20, 10), 0.5) }.play;`
and it worked fine.


danstowell:
I am not 100% sure, but I don't think you should simply be deleting the "load" lines.

I remember a while back, someone implemented the `PluginLoad()` macro specifically to allow compiling on MSVC, because of the name clash with `load`. I think what you should do is replace the "load" lines with `PluginLoad(pluginname)`. I think on Windows this then uses a modified function name [as seen here on this line](https://github.com/supercollider/supercollider/blob/master/include/plugin_interface/SC_InterfaceTable.h#L224).

timblechmann:
removing the 'load' declarations is fine. `PluginLoad` does the following

* generate explicit names for the load functions to statically link plugins into scsynth (smells iOSy)
* generate a function for server ABI versioning
* generate a function to allow selection of plugin for supernova or scsynth.

in the past it used a modified function name, but only because the `load` symbol wasn't exported correctly. plugins should already use `PluginLoad`, otherwise they'd crash the server depending on which abi they are compiled for.

--

in general, all looks good to me, though `xor` could be pulled in via `#include <iso646.h>` and `std::min` might need some type hints as both arguments are expected to have the same type.

llloret:
Hi, @danstowell, the code already had `PluginLoad`s, that's why MSVC was complaining, because it was seeing the load AND the PluginLoad. Also, from what I can see the compilation was fine on Travis.

danstowell:
Thanks both. Fine by me then.

( @llloret note that the Travis build only checks that things can build. It doesn't check that the plugins run...)

bagong:
So we have blessing, that's great, thank you two! @lorret: I guess now it has boiled down to the real build issues, I still get 20:

````
"C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\install.vcxproj" (default target) (1) ->
"C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\ALL_BUILD.vcxproj" (default target) (3) ->
"C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JPverb.vcxproj" (default target) (33) ->
(ClCompile target) ->
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\DEINDUGens\JPverbRaw.cpp(581): fatal error C1061: compiler limit : blocks nested too deeply [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\so
urce\JPverb.vcxproj]


"C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\install.vcxproj" (default target) (1) ->
"C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\ALL_BUILD.vcxproj" (default target) (3) ->
"C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.vcxproj" (default target) (36) ->
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(1001): error C2057: expected constant expression [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.
vcxproj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(1001): error C2466: cannot allocate an array of constant size 0 [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\sour
ce\JoshPVUGens.vcxproj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(1001): error C2133: 'magarray' : unknown size [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.vcx
proj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(1030): error C2057: expected constant expression [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.
vcxproj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(1030): error C2466: cannot allocate an array of constant size 0 [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\sour
ce\JoshPVUGens.vcxproj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(1030): error C2133: 'magarray' : unknown size [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.vcx
proj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(1978): error C2057: expected constant expression [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.
vcxproj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(1978): error C2466: cannot allocate an array of constant size 0 [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\sour
ce\JoshPVUGens.vcxproj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(1978): error C2133: 'bins' : unknown size [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.vcxproj
]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(2071): error C2057: expected constant expression [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.
vcxproj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(2071): error C2466: cannot allocate an array of constant size 0 [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\sour
ce\JoshPVUGens.vcxproj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(2071): error C2133: 'bins' : unknown size [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.vcxproj
]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(2312): error C2057: expected constant expression [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.
vcxproj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(2312): error C2466: cannot allocate an array of constant size 0 [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\sour
ce\JoshPVUGens.vcxproj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(2312): error C2133: 'bins' : unknown size [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.vcxproj
]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(2402): error C2057: expected constant expression [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.
vcxproj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(2402): error C2466: cannot allocate an array of constant size 0 [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\sour
ce\JoshPVUGens.vcxproj]
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshPVUGens.cpp(2402): error C2133: 'bins' : unknown size [C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\JoshPVUGens.vcxproj
]


"C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\install.vcxproj" (default target) (1) ->
"C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\ALL_BUILD.vcxproj" (default target) (3) ->
"C:\Users\Rainer\Projects\sc\sc3-plugins\build_MW_Rl\source\MCLDGetenvUGen.vcxproj" (default target) (47) ->
  C:\Users\Rainer\Projects\sc\sc3-plugins\source\MCLDUGens\MCLDGetenvUGen.cpp(7): fatal error C1083: Cannot open include file: 'sys/time.h': No such file or directory [C:\Users\Rainer\Projects\sc\
sc3-plugins\build_MW_Rl\source\MCLDGetenvUGen.vcxproj]
`````

Do you have time to look into them? I guess the next step, testing cross-platform that things actually work would make most sense after the build has succeeded? Is that the right order?

**Comment by danstowell** [comment]
To fix the issue in MCLDGetenvUGen.cpp: on windows, don't attempt to include 'sys/time.h'. Instead, it looks like you can include `/include/common/wintime.h` - I've never used that myself

**Comment by llloret** [comment]
Yes, I will look into those. 

For the `JPverbRaw`, it is a MSVC bug (https://support.microsoft.com/en-us/kb/315481). So, basically every for in that function needs to be enclosed in braces. 

For `JoshPVUGens.cpp` I think they need to be allocated dynamically, not statically as it is doing now.

`MCLDUGens`, is using getTimeOfDay, which is not implemented in Windows. I think I may have a workaround for it.

So, basically, I think I understand what needs to be done in them. I will try to finish them as soon as I can.

**Comment by bagong** [comment]
Fab! Thank you very much!

**Comment by llloret** [comment]
Hi, @bagong, can you restart the Travis build when you can. I am not sure why it failed, but I don't think it was because of a compilation error.
Thanks

**Comment by gusano** [comment]
@llloret Just restarted it

**Comment by llloret** [comment]
@gusano, thanks.

**Comment by llloret** [comment]
Hi, guys, I am not sure if I am going to have time to work on fixing the Clockmus Unit in MCLDGetenvUGen.

For Windows, would it be ok to put a pragma warning saying that it is not ready yet, and disable it for now, but make the rest of the plugins available?

Just to keep Windows appealing, since at the moment no sc3-plugin compiles with MSVC...

let me know what you think.

**Comment by bagong** [comment]
I think it depends on how much time you have in midterm, ie. for example during the next month. There is no reason for haste, I am sure it will take another month before 3.7.2 hit's the ground. So far the main reason for doing another 3.7 release would be your changes anyways. And we do have the option of using MinGW built 64-bit plugins, we haven't had problem reports so far... Building the sc3-plugins with VS is most useful if we plan to switch to VS as only build environment for Windows. But then supernova would also have to be made compatible. If you ask me, I would prioritize good MIDI support above sc3-plugins. And no need for haste on either of them. It would just be good to know if you are committed to go on after some time (let's say within a few weeks) or not. It's up to you of course!

As to the sys/time.h not found error. I've seen that in main SC as well, maybe there is an implementation that could be used as model? Time is a recurring problem in the Windows build. Maybe somebody has a suggestion how to do that? Maybe @sofakid or @timblechmann ?

**Comment by bagong** [comment]
PyrSched.cpp uses gettimeofday and includes "wintime.h" for Windows. I think Lucas Cornelisse (@sofakid) wrote wintime.h when he was working towards the VS 3.7 build. But I also see a lot of references to timeofday in boost, but the Windows references seem to require MinGW?

**Comment by llloret** [comment]
I have something that should work using the windows Api, but when I try to
compile, I am getting problems in Winnt.h, because it says that IN is not
defined...

**Comment by llloret** [comment]
oops!, sorry, I didn't trim the original message when I replied... naughty
boy. Sorry about that...

>

**Comment by llloret** [comment]
Ok, I think I have it. The problem was the SC_Plugin.h undefines the macro IN, which is used in winnt.h. Moving the #include down fixed the issue.

I will check that it is ok, and commit.

**Comment by llloret** [comment]
@bagong, thanks for the top about `wintime.h`, it was similar to what I was doing, so it has saved some code duplication. After including that, and changing the order of the includes, it compiled nicely.

Just note that I have not tested the plugins, except for the Blip one. I have only made changes so that it compiles with Windows, so there might be something wrong in there. This way people can chip in, and start giving it a go before we put into master. I will try to do some testing, anyway.

I will move back to MIDI now, and see if I can advance some more there.

**Comment by bagong** [comment]
Fantastic, thank you again. I will spend time tomorrow with the plugins!

**Comment by llloret** [comment]
Hi, @bagong, I know that you commented something about a couple of plugins that were not working fine, but I cannot find where you made that comment. Would you mind putting it here, for reference? I could have a quick look, although if there are Faust related as you were saying, I won't be able to do much,

Also, are those the only problems that you have found, or are there more? I am not sure if you were able to try all of the plugins, or you are still going through them... there were lots :).

Thanks

**Comment by bagong** [comment]
Yea, sorry @llloret , there are some badly placed elements in our threads. It was in one of the midi issues:

This I get on server boot:

*** ERROR: API Version Mismatch: C:\Users\Rainer\AppData\Local\SuperCollider\Extensions\SC3plugins\Greyhole.scx
*** ERROR: API Version Mismatch: C:\Users\Rainer\AppData\Local\SuperCollider\Extensions\SC3plugins\JPverb.scx

This might somehow be Faust related.

**Comment by llloret** [comment]
Thank you, @bagong.

**Comment by llloret** [comment]
I have just committed a couple of fixes that should make this work under Windows. I have checked that they compile and load ok. @bagong, could you give them a try, and see if they work ok?

As fare as I understand this was the last outstanding issue to make the sc3-plugins compile with MSVC. Let me know if there is something else.

**Comment by sonoro1234** [comment]
@llloret  I have seen you use a lot of 
```
#ifdef _WIN32
```
when you meant MSVC (not mingw) so all of this will break compilation for mingw

**Comment by sonoro1234** [comment]
could be
```
#ifdef _MSC_VER
```

**Comment by bagong** [comment]
This is an interesting point, @sonoro1234 . So far it has *not* broken the *build*. But it might have changed the resulting binaries a bit. I just built the very latest version with MinGW like in the 3.7.1 release, using MinGW 4.9.2. I guess the changes could be both advantageous or disadvantageous. Keeping the codebase of both Windows compiles as close as possible should make the Windows version easier to maintain, but it might occasionally be disadvantageous for either of them. Maybe you could give some specific points you have in mind in the current changes?

File: source/JoshUGens/JoshPVUGens.cpp [hunk 1]
```diff
@@ -966,6 +967,21 @@ void PV_MagMap_next(PV_MagMap* unit, int inNumSamples)
 
 
 /* a function for sorting floats */
```
sonoro1234: dont use this solution on mingw

joshpar: I'm confused - is there something I need to do here?
Best
Josh

/*
Josh Parmenter
www.realizedsound.net/josh
*/

sonoro1234:
the only bad one I get is on  source/JoshUGens/JoshPVUGens.cpp 
we should avoid vector whenever possible because it uses not RT memory

source/JoshUGens/JoshPVUGens.cpp
bagong: I suspect you have already built :D. In fact there are warnings in the MinGW build of the JoshUGens (not sure if they are new though, there have been quite a few of them in the past too):

`````
[ 25%] Building CXX object source/CMakeFiles/JoshGrainUGens.dir/JoshUGens/JoshGrainUGens.cpp.obj
C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshGrainUGens.cpp:6739:0: warning: "GRAIN_BUF" redefined
 #define GRAIN_BUF             \
 ^
C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshGrainUGens.cpp:731:0: note: this is the location of the previous definition
 #define GRAIN_BUF \
 ^
C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshGrainUGens.cpp:6815:0: warning: "GRAIN_LOOP_BODY_4" redefined
 #define GRAIN_LOOP_BODY_4          \
 ^
C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshGrainUGens.cpp:746:0: note: this is the location of the previous definition
 #define GRAIN_LOOP_BODY_4 \
 ^
C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshGrainUGens.cpp:6844:0: warning: "GRAIN_LOOP_BODY_2" redefined
 #define GRAIN_LOOP_BODY_2        \
 ^
C:\Users\Rainer\Projects\sc\sc3-plugins\source\JoshUGens\JoshGrainUGens.cpp:776:0: note: this is the location of the previous definition
 #define GRAIN_LOOP_BODY_2 \
 ^
``````

File: source/MCLDUGens/MCLDGetenvUGen.cpp [hunk 2]
```diff
@@ -3,9 +3,13 @@ Getenv UGen (c) Dan Stowell 2007
 Grab environment variables (numeric ones please!) for use in synth graphs.
 */
 
-#include "SC_PlugIn.h"
+#ifdef _WIN32
```
sonoro1234: _MSC_VER

bagong: Could you elaborate a bit? We have a similar case in SC proper, would be good if both places were treated the same.

source/JoshUGens/JoshPVUGens.cpp
bagong: @joshpar , would be great if you could comment on the intended changes...

source/MCLDUGens/MCLDGetenvUGen.cpp
sonoro1234: I mean use _MSC_VER instead of _WIN32

bagong: Well *that* I understood ;) Could you explain a bit why that is disadvantageous, as you did for the changes concerning Joshs' UGens? But I just saw, there is only one case in PyrSched.cpp, and that in fact uses _MSC_VER. Thanks!

sonoro1234: Well, a general reason would be: If it is ok for gcc and works also with mingw leave it as it is and only change that for MSVC.

llloret:
Ok, so what should we do with ` _WIN32` vs `_MSC_VER`? You are raising a good point...

@sonoro1234, regarding the `vector`, I totally agree with you, but since MSVC does not support variable length arrays, and I didn't want to tinker with the code too much, I left like that.

Anyway, it would be interesting to see what the gcc compiler does when it sees a variable length array in the stack (perhaps it does something similar to initializing a vector?).

llloret:
I think that for the sc3-plugins it probably makes sense to use _MSC_VER as you suggest. Not so sure about SC itself.

sonoro1234:
I would always use _MSC_VER with everything that is already working with mingw.
Of course is it does not work with mingw we could group the windows solutions under _WIN32

llloret:
@bagong, let me change the sc3-plugins to use _MSC_VER, and commit again.

We should review the usage on SC as well.

bagong:
Sure!

source/JoshUGens/JoshPVUGens.cpp
```diff
@@ -968,6 +967,21 @@ void PV_MagMap_next(PV_MagMap* unit, int inNumSamples)
...
+#ifdef _MSC_VER
+bool isfloatgreater(float a, float b);
+bool isfloatgreater(float a, float b)
+{
+	return (a < b);
+}
+
+
+bool isfloatless(float a, float b);
+bool isfloatless(float a, float b)
```
timblechmann: `std::less<float>` would be the c++ way

source/JoshUGens/JoshPVUGens.cpp
timblechmann: variable sized arrays are a gcc extension to c++, clang emulates them, but they are not standard-compliant.

two possibilities:
* over-allocate a `std::array<>` with fixed-sized limit
* `alloca` or `_Alloca`

source/JoshUGens/JoshPVUGens.cpp
```diff
@@ -1009,7 +1028,11 @@ void PV_MaxMagN_next(PV_MaxMagN* unit, int inNumSamples)
 
 	float numpars = IN0(1);
 
+#ifdef _MSC_VER
+	std::sort(magarray.begin(), magarray.end(), isfloatless);
```
timblechmann: compare: http://en.cppreference.com/w/cpp/algorithm/sort

`std::sort` sorts via `operator<` by default

source/JoshUGens/JoshPVUGens.cpp
```diff
@@ -1038,7 +1066,11 @@ void PV_MinMagN_next(PV_MinMagN* unit, int inNumSamples)
 
 	float numpars = IN0(1);
 
-	qsort(magarray, numbins, sizeof (float), isfloatgreater);
+#ifdef _MSC_VER
+	std::sort(magarray.begin(), magarray.end(), isfloatgreater);
+#else
+	qsort(magarray, numbins, sizeof(float), isfloatless);
```
timblechmann: `isfloatless`?

source/JoshUGens/JoshPVUGens.cpp
```diff
@@ -1038,7 +1066,11 @@ void PV_MinMagN_next(PV_MinMagN* unit, int inNumSamples)
 
 	float numpars = IN0(1);
 
-	qsort(magarray, numbins, sizeof (float), isfloatgreater);
+#ifdef _MSC_VER
+	std::sort(magarray.begin(), magarray.end(), isfloatgreater);
```
timblechmann: `std::sort` with `std::greater<>`

http://en.cppreference.com/w/cpp/utility/functional/greater

source/DEINDUGens/GreyholeRaw.cpp
```diff
@@ -49,7 +49,7 @@ using namespace std;
 #if defined(__GNUC__) && __GNUC__ >= 4
 # define FAUST_EXPORT __attribute__((visibility("default")))
 #else
-# define FAUST_EXPORT /* NOP */
+# define FAUST_EXPORT extern "C" SC_API_EXPORT
```
timblechmann: maybe file a bug report at faust to properly support msvc?

source/JoshUGens/JoshPVUGens.cpp
llloret: Yep, I introduced this error. I have fixed it now.

bagong:
Are there remaining reservations? We need to commit this for the "imminent" release ;)

source/DEINDUGens/GreyholeRaw.cpp
llloret: @timblechmann Faust has accepted the change upstream

LFSaw:
just for future reference, please _do not_ manually change FAUST-generated `.cpp`-files (as JPVerbRaw.cpp in DEINDUGens) since those are, well, generated.

PR #87, opened 2016-05-30
BlitB3 -- fix oscillator freezing

### LINKED ISSUE #82: BlitB3 can "freeze" at low frequencies
The BlitB3 uGen is freezing on Windows 10 (64-bit) using the pre-compiled plug-in library and SC 3.7.1 (also 64-bit).

I'm using the following line of code:

`{BlitB3.ar(MouseX.kr(0.01, 10000))}.scope`

If I move my mouse to the far left of my screen, the uGen will occasionally stop generating. Sometimes, it will start generating again after a few seconds. I did one test where it took about two minutes to start generating again.

snappizz:
Fixes #82 with smarter phase tracking. A few extra divides and multiplies per control period, but shouldn't be huge.

Sorry about the absurd diff, my editor removed a bunch of trailing whitespace. I didn't feel like turning it off because we might as well get rid of it.

### DIFF
--- a/source/AntiAliasingOscillators/AntiAliasingOscillators.cpp
+++ b/source/AntiAliasingOscillators/AntiAliasingOscillators.cpp
@@ -93,6 +93,10 @@ struct DPW3Tri : public Unit
 	double differentiations_[2];
 };
 
+struct BlitB3D : public Unit {
+	float phase;
+};
+
 
 extern "C" {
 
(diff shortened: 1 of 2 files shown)


telephon:
looks good. also thanks for cleaning the code!

bagong:
How about merging this for 3.7.2, i.e. very soon? ;)

snappizz:
Hmm, reading some recent discussions on music-dsp, using floating point to keep track of phase increases the risk of phase drift.

PR #95, opened 2016-06-14
convert help files to SCDoc format

dotinspace:
Converted multiple html help files to SCDoc format.


danstowell:
Good work @dotinspace! One thing - shouldn't this PR also delete the old HTML files that are getting replaced, as well as adding new files?

dotinspace:
Yeah, @danstowell, that makes sense. Done.

snappizz:
33/71

PR #96, opened 2016-06-16
JoshUgens: help: update to SCDoc

dotinspace:
Converted more of the old html help files to SCDoc format.


snappizz:
44/71!

dotinspace:
Actually, 45/71. BFMixer is done, too.

PR #98, opened 2016-06-18
joshugens: help: convert from old html to new SCDoc format

dotinspace:
Converted old html help files to new SCDoc format, removed old help
files, fixed description of `PVAna.schelp`.


snappizz:
66/71!!!

PR #99, opened 2016-06-19
joshugens: converted help files & uncommented Tendency.sc

dotinspace:
Converted the rest of the old help files, excluding DelTapRd/Wr, and
deleted unneeded html help files. Also, uncommented the Tendency.sc
class source, so that the examples in Tendency help file are useable.
So, that's it- done! 71/71 old help files converted.


joshpar:
Actually, I think Tendency was moved to a quark. I'll double check, but I'm guessing those files can just go away.
Thanks for all the work!
Josh

/*
Josh Parmenter
www.realizedsound.net/josh
*/

snappizz:
yes, Tendency is now at https://github.com/supercollider-quarks/Tendency.

PR #108, opened 2016-08-15
DWGUGens: fix float/double confusion for GCC 6 compatibility

### LINKED ISSUE #106: source/DWGUGens/dwglib/DWG.hpp fail to build with GCC 6
The source fail to build in Debian after Debian switched to GCC 6 as the default C++ compiler.  See https://bugs.debian.org/831078 for details.

The problem is that this expression no longer work in source/DWGUGens/dwglib/DWG.hpp:
```
float loss;
...
void set_area(float a1,float lossF)
    {
        if(a1 < 1e-18)
            loss = 0.0;
        else
            loss = 1.0 - std::min(lossF/sqrt(static_cast<double>(a1)),1.0);

    }
```

I believe the cause is that the compiler now pick the function float sqrt(float) instead of the function double sqrt(double).  There are two ways to solve the build problem, and I am unsure which one is the best.  Either replace 1.0 with 1.0f to use float everywhere in the calculations, or cast a1 using static_cast(double>(a1) to tell the compiler to use doubles everywhere.  I assume the former calculation is slightly quicker, while the latter is more accurate.

Do you have an opinion on how this should be solved?  How important is the accuracy of the calculation?

snappizz:
Fix #106 

I can't reproduce the original bug but I can confirm that this doesn't cause any problems

### DIFF
--- a/source/DWGUGens/dwglib/DWG.hpp
+++ b/source/DWGUGens/dwglib/DWG.hpp
@@ -22,7 +22,7 @@
 #include "SC_PlugIn.h"
 extern  InterfaceTable *ft;
 #include <float.h>
-#include <math.h>   
+#include <math.h>
 #ifndef M_PI
 #define M_PI           3.14159265358979323846
 #endif
@@ -53,7 +53,7 @@ void volcar(const char *_Message, const char *_File, unsigned _Line);
 #define b2(x)   (   (x) | (   (x) >> 1) )
 #define b4(x)   ( b2(x) | ( b2(x) >> 2) )
 #define b8(x)   ( b4(x) | ( b4(x) >> 4) )
-#define b16(x)  ( b8(x) | ( b8(x) >> 8) )  
+#define b16(x)  ( b8(x) | ( b8(x) >> 8) )
 #define b32(x)  (b16(x) | (b16(x) >>16) )
 #define next_power_of_2(x)      (b32(x-1) + 1)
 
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
Note, this is the error from gcc 6:

```
In file included from /home/pere/src/sonic-pi/pkg-supercollider-sc3-plugins/source/DWGUGens/dwglib/DWG.cpp:19:0,
                 from /home/pere/src/sonic-pi/pkg-supercollider-sc3-plugins/source/DWGUGens/DWGPlucked.cpp:19:
/home/pere/src/sonic-pi/pkg-supercollider-sc3-plugins/source/DWGUGens/dwglib/DWG.hpp:1271:53: note:   mismatched types 'std::initializer_list<_Tp>' and 'float'
             loss = 1.0 - std::min(lossF/sqrt(a1),1.0);
                                                     ^
```

This pull request changes the wrong 1.0 value.  The problem is the arguments to std::min() and a fix is to replace the last 1.0 with 1.0f.

sonoro1234:
@snappizz 
Thanks for taking care. I dont know how to continue this. Perhaps is better that you finish that instead of me for not overlapping PRs.
```
loss = 1.0 - std::min(lossF/sqrt(a1),1.0f);
```

snappizz:
according to @petterreinholdtsen's comment, this PR is wrong so I'll start over.

snappizz:
ok, #112 should do it.

PR #109, opened 2016-08-18
Build plugins for supernova if we built it

vivid-synth:
We should at least be building plugins for supernova when we're building it in the SC repo

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
 cmake_minimum_required (VERSION 2.8)
+set(SUPERNOVA_CMAKE_MINVERSION 3.1)
 project (sc3-plugins)
 
 if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/external_libraries/nova-simd/simd_memory.hpp)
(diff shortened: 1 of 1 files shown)


danstowell:
I don't object - but I'd just point out that this is effectively "guessing" that supernova is available/wanted. (Your comment says "if we built it" or "when we're building it" but that's not actually tested for.)

vivid-synth:
@danstowell true, but it's the same cmake test we use for building supernova by default. I'd be all for compiling it all the time (or anytime supernova is installed) if that's possible, but I at least don't want to be building supernova by default but with no ugens to use it with.

danstowell:
"no ugens" isn't strictly true now is it? but np, go ahead

vivid-synth:
@danstowell -- oh, very true: you mean the ones from the SC repo, right?

Also, I've only done the flag this way because of my limited knowledge of cmake: is there a way to say "if supernova is installed"?

danstowell:
Well you could check for the existence of the executable `supernova` in the path, but meh, that feels a bit flakey too. (Supernova doesn't install headers of its own; if it did, I'd say "check for those")

bagong:
No prob for Windows.

PR #112, opened 2016-08-28
DWGUGens: fix float/double confusion for GCC 6 compatibility (take 2)

### LINKED ISSUE #106: source/DWGUGens/dwglib/DWG.hpp fail to build with GCC 6

snappizz:
Fix #106.

### DIFF
--- a/source/DWGUGens/dwglib/DWG.hpp
+++ b/source/DWGUGens/dwglib/DWG.hpp
@@ -22,7 +22,7 @@
 #include "SC_PlugIn.h"
 extern  InterfaceTable *ft;
 #include <float.h>
-#include <math.h>   
+#include <math.h>
 #ifndef M_PI
 #define M_PI           3.14159265358979323846
 #endif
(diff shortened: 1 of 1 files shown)


sonoro1234:
With sc_min instead of std::min this seems not neccessary: see #106

bagong:
Can't test higher than gcc4.8, but no prob on Windows.

PR #114, opened 2016-09-21
EnvFollow: initialize outputs

snappizz:
For background: https://github.com/supercollider/supercollider/issues/2333. Sorry if it's kind of weird to only fix this one when the rest of the SLUGens are also in need of attention. I use EnvFollow a lot.


bagong:
No prob on Windows.

crucialfelix:
lgtm

PR #123, opened 2016-11-05
VBAP: interpolate panning arguments

snappizz:
Right now my sc3-plugins development environment is not functional, so I can't test this. I can only tell that it compiles.

### DIFF
--- a/source/VBAPUGens/VBAP.cpp
+++ b/source/VBAPUGens/VBAP.cpp
@@ -19,6 +19,8 @@
  beastmulch-info@contacts.bham.ac.uk
 
  The BEASTmulch project was supported by a grant from the Arts and Humanities Research Council of the UK: http://www.ahrc.ac.uk
+
+ Smoother modulation implemented by Nathan Ho, 2016.
 */
 
 
@@ -95,6 +97,10 @@ static InterfaceTable *ft;
 
 struct VBAP : Unit
 {
+	float last_azimuth;
+	float last_elevation;
+	float last_spread;
+
 	float x_azi;				/* panning direction azimuth */
 	float x_ele;				/* panning direction elevation */
 	float x_set_inv_matx[MAX_LS_SETS][9];  /* inverse matrice for each loudspeaker set */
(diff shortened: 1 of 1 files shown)


patrickdupuis:
@snappizz what are you thoughts on this today?

I've been able to compile this. Was able to run a VBAP example successfully (although i'm not equipped to test multichannel speaker arrays).

muellmusik:
I can test it.

muellmusik:
I've also got a version that's updated to VBAP-1.0.3.2 and has another fix from Luca. Would be good to get those in.

muellmusik:
Two questions:

1. Given that channel gains are already interpolated, do you have an example which demonstrates the audible difference of this PR?
2. As VBAP is not really cheap, would a better approach not be to provide next funcs which allow for ar inputs? This would leave the user in control.

snappizz:
i don't remember anymore what i was doing here. [this thread might be helpful](http://new-supercollider-mailing-lists-forums-use-these.2681727.n2.nabble.com/VBAP-distortion-td7628552.html). sorry, i'm spending my SC time focusing on other things right now and haven't really thought about this.

patrickdupuis:
@muellmusik have you had a chance to test? There are a few people I've met this week that depend on VBAP (notably for this project https://gitlab.com/sat-metalab/SATIE).

snappizz:
there's no need to get an actual multichannel speaker system to test this UGen (and beside the noise can be a little hard to hear) -- the bug is clearly visible via plotting. i'll have time to work on this PR tomorrow.

muellmusik:
Sorry, I've not had a moment. I won't have a lot of time for the next two weeks.

> the bug is clearly visible via plotting.

Please could we have a reproducer and a clear description of the problem on this PR? That way we'll know exactly what we're trying to fix and can test different branches or approaches.

snappizz:
sorry for being evasive for so long about the actual problem at hand.... now that i have some time to work on this, here's the reproducer:

    ~array = VBAPSpeakerArray(2, [-30, 30, 0, -110, 110]);
    ~buf = ~array.loadToBuffer;
    { VBAP.ar(5, DC.ar(1), ~buf.bufnum, Line.kr(0, 360, 0.05), 1.0, 0.0) }.plot(0.05);
    { VBAP.ar(5, DC.ar(1), ~buf.bufnum, Line.ar(0, 360, 0.05), 1.0, 0.0) }.plot(0.05);

the last two lines are simply panning a DC signal around the array. the problem is that the gain of each speaker is modulating at control rate, and stairstepping -- even when the azimuth signal is audio rate:

vbap

this results in a noisy buzz right at whatever the control rate frequency is. the expected behavior would be to accept an audio-rate azimuth signal, and to interpolate linearly if the signal is control rate.

this PR kinda fixes it, but it looks like i've somehow managed to zero out the gain at every control period:

vbap2

so that's the status here, i'm going to work on this now.

snappizz:
#183

PR #124, opened 2016-11-05
Warn if MathLib is not installed

snappizz:
super quick fix for 3.8

### DIFF
--- a/source/ATK/sc/Classes/ATK.sc
+++ b/source/ATK/sc/Classes/ATK.sc
@@ -114,6 +114,11 @@ Atk {
 
 		// supported sets
 		sets = [ 'FOA', 'HOA1', 'HOA2', 'HOA3', 'HOA4', 'HOA5'];
+
+		if (\Matrix.asClass.isNil) {
+			"Please run Quarks.install(\"MathLib\") and recompile the class library."
+			" It is a dependency for the Ambisonic Toolkit (ATK).".warn;
+		};
 	}
 
 	*userSupportDir_ {arg userSupportDirIn;


bagong:
Looks good and tried on OSX and Windows.
Can we ignore that empty merge of master and commit?
Thanks!!

snappizz:
done

crucialfelix:
Sorry, not following something here. This PR posts a warning in *initClass

But the problem is that there is an extension to Matrix: https://github.com/supercollider/sc3-plugins/blob/e46f54d8e6b4ea399bc4f9b9d7aa640923146089/source/ATK/sc/Classes/extMatrix.sc

I forget - does sc just warn when it sees an extension to a non-existing class ? Does it still compile ?

bagong:
It does still compile, but the error message gives the user no indication how to solve the problem.
The Matrix extension comes with ATK, so the class initialization seems like a reasonable place to make the check? No?

telephon:
So this basically makes MathLib a dependency for sc3-plugins. Somehow this isn't optimal. But I suppose it is not acceptable to make the Ambisonic toolkit itself a Quark, while its UGens are in the plugins …

snappizz:
@telephon see [this thread on sc-dev](http://new-supercollider-mailing-lists-forums-use-these.2681727.n2.nabble.com/Fwd-sc-users-Re-SC3-plugins-missed-class-td7628763.html). right now this is the best solution we can get under our current time constraints.

jamshark70:
Felix proposed, on the mailing list:

~~~
Agreed - that should be a method on AtkMatrix. 

Actually it doesn't need to exist at all. It's just taking a normal Matrix, converting it to an AtkMatrix and then saving it.

AtkMatrix.newFromMatrix(matrix).writeToFile(fileNameOrPath, set, type, note, attributeDictionary, overwrite);

is fine IMO.

versus:

matrix.writeToFile(fileNameOrPath, set, type, note, attributeDictionary, overwrite);
~~~

This should be feasible under the time constraints, and IMO eliminating the dependency is a better solution.

jamshark70:
I will reiterate my concern that many, MANY people use sc3-plugins without using the ATK classes specifically. These users should not be bothered.

I guess if I had to state it as a principle, it would be, confine the damage to the smallest area. Felix's suggestion would break code for some specific ATK uses, but I've played a little with ATK and you can do a lot with it and never (as a user) call `writeToFile` directly. If it's called internally, that could be fixed. As far as I can see, this is a relatively small impact.

By contrast, this approach suggests to *all* sc3-plugins users that they must install an extra quark (and if they don't already have git, they will get errors with this)... and in fact, it's optional if they are not using ATK but the warning doesn't say so.

IMO Felix's idea is better and it's unfortunate that it was overlooked.

crucialfelix:
Nathan took a look and said that it's a core part of the API.  So not so
simple to remove it.

snappizz:
It's not really a core part of the API, but it is public-facing. It's only been around since September, however, so it's probably not so bad to retract.

bagong:
So from a purely organizational perspective, may I suggest the following: a) We agree that no binaries will be changed for 3.8, then Sonic Pi can use the current state for their release.  b) As to SC, I'll upload Windows beta2 builds on Wednesday, containing this PR, if no final solution has been agreed upon until then. I won't be available for a while after that. If anybody else can provide builds, that's fine of course, otherwise I'll upload the release version in January.

jamshark70:
> It's not really a core part of the API, but it is public-facing. It's only been around since September, however, so it's probably not so bad to retract.

Ok, thanks for the clarification. I realize I'm contradicting my position about not breaking code, but I think it's more that not breaking code is often the best way to minimize damage... but it isn't *always* the least amount of damage.

It doesn't affect me because I have MathLib anyway, but I'm thinking of the new user who installs SC and sc3-plugins for the first time, never used SC before, and the first thing she sees is "OMG you're missing something and you're just going to die if you don't have it," and then she tries to run the Quarks command and it complains that it can't find git, so she says, ok, back to Max. It is not a good first experience.

snappizz:
#125

vivid-synth:
In terms of depending on MathLib, how does that interact with the ```-DSC_SCLANG=False``` (https://github.com/supercollider/supercollider/pull/1939) flag?

crucialfelix:
It only depends on Matrix because it adds a class extension to it. The plug
ins themselves are not dependent on mathlib.

joslloand:
Hi All, You'll see the commit I just made. This is the solution @mtmccrea has contributed. MathLib is removed as a dependency. The extension for writing that the ATK now adds is to Array, part of the core.

For the longer term we're thinking that moving all the language side elements to a Quark is probably the best idea. (We've been considering something along those lines for future development.) IN terms of the current release cycle, doing so now seems to be somewhat risky.

PR #125, opened 2016-11-06
ATK: remove Matrix:writeToFile

snappizz:
It is probably unceremonious to simply remove the section on writing raw matrix files, but I am not knowledgeable enough to fix it and too lazy to figure it out myself.


bagong:
Thank you for your initiative and attempt to do something @snappizz , that's really kind, especially as this is not your fault. Now the sclang boot error is gone, but ATK is considerably broken. MathLib is not only required to write files, half of "help/Guide to ATK Matrix Files" is broken. It's a pity neither @joslloand, @dmartinp or @mtmccrea seem aware of our release schedule and the resulting urgency of the issue.

snappizz:
I feel that maybe you're being a bit too hard on the ATK devs, but with this whole situation, it's clear that ATK is really a project of its own, separate from sc3-plugins. [As I proposed on the dev list](http://new-supercollider-mailing-lists-forums-use-these.2681727.n2.nabble.com/Fwd-sc-users-Re-SC3-plugins-missed-class-td7628763.html), it may be best to move ATK out -- not because ATK is harmful to sc3-plugins, but because sc3-plugins is weighing down ATK development.

snappizz:
@bagong The characterization of ATK being "broken" all of a sudden is unfair, since it's depended on MathLib for a while (since 3.7 at least). What's new is that failure to fulfill the dependency now creates a scary-looking error *on startup*.

bagong:
Sorry, @snappizz, I was just trying to say that this PR breaks more than what you removed from help. I am just trying to describe what I see, not criticize anybody. Actually just hoping that the ATK contributors come to our help.
Btw, would it maybe be an option to add that Matrix extension to MathLib?

snappizz:
> this PR breaks more than what you removed from help

Huh? Matrix:writeToFile is more or less a wrapper around AtkMatrix:writeToFile. The rest of the help file concerns the latter function, which is still intact. So it's fine.

> Btw, would it maybe be an option to add that Matrix extension to MathLib?

It depends on ATK, so no. It wouldn't create an error on startup, but it wouldn't work on its own either.

mtmccrea:
@bagong - just as a heads up, we are aware of the issue and working on it.  

> It's a pity neither @joslloand, @dmartinp or @mtmccrea seem aware of our release schedule and the resulting urgency of the issue.

In fact we've reached out multiple times to coordinate with the 3.8 release, this dependency issue with Matrix was simply a mistake and we plan to fix it ASAP. While it may be a small fix (felix's proposed solution wouldn't actually work, unfortunately), the ATK is coordinated through multiple people (which happens outside of the user/dev list) and so even small changes require at least a couple people to review it.  Apologies that this is holding up the release.

@snappizz - thanks for your thoughts on ATK's move to being self-contained. @joslloand and I have had many discussions about this, which we will likely bring to the list when we have the time... we'd very much like to move it from SC3Plugins!

bagong:
@mtmccrea , thanks for your answer! I think the underlying problem is that SC because of missing wo-manpower cannot guarantee a well coordinated and regular release cycle, atm and that therefore communicational mishaps are likely to occur. I am sorry (also to @snappizz ) if my comment above sounded rude, it was not my intention at all and is maybe due more to language deficiencies than anything else. I personally like that sc3-plugins contains such a classy and well maintained extension. Midterm it might be nice to move the classes to a Quark, but that's just my personal opinion ;)
@mtmccrea, do you think you could provide a fix until Wednesday? That would allow me to provide proper Windows builds before I have to leave for quite a while.

jamshark70:
> MathLib is not only required to write files, half of "help/Guide to ATK Matrix Files" is broken.

Isn't that a documentation issue? If ATK installation instructions tell users to install the MathLib quark, wouldn't that take care of it?

It's consistent with my concern elsewhere: the users who should be concerned with ATK dependencies should be people actually using ATK, and not all users.

jamshark70:
Also, I don't think it would be objectionable to leave the ATK UGen c++ code in sc3-plugins and move the classes to a quark. Then the quark could name MathLib as a dependency and, bang, problem solved.

mtmccrea:
I've submitted a fix for the current Matrix dependency issue to our local repo for review, so I expect the patch will be along shortly. Thanks for your patience with this.

> Also, I don't think it would be objectionable to leave the ATK UGen c++ code in sc3-plugins and move the classes to a quark.

This is one of the options we're considering. @snappizz had also suggested a modification of the Quarks system that would allow server plugins, which would be even better. I did a basic test a while back that simply place the plugins alongside the class files in a quark. It pulled from the repository just fine, but SC couldn't load plugins that were in the Extensions/downloaded-quarks directory.

The auto-building of plugins with travis is a convenience that would be lost if we moved the plugins to Quarks, but on the other hand, getting travis to build individual plugins in disparate repositories is another project (which I had [working in a basic way](https://github.com/mtmccrea/atk-plugins-build-test), but would take work to generalize).

Again, apologies for any delays this has caused the 3.8 release! We appreciate the work that everyone has put into it!

vivid-synth:
One thing that would be nice to keep in mind (although it's a bit blue-sky at the moment) is to keep the UGens sufficiently decoupled from the classes that other, non-sclang clients could build tools to to Ambisonics work. This could be by keeping the UGens in sc3-plugins, or just by being mindful of the dependencies if/when they're moved to their own project/Quark. I don't have the time at the moment to do the development but I'd love to do some Ambisonics work in Haskell, for example.

> @snappizz - thanks for your thoughts on ATK's move to being self-contained. @joslloand and I have had many discussions about this, which we will likely bring to the list when we have the time... we'd very much like to move it from SC3Plugins!

Since there are a lot of moving parts, it would be great to bring your thoughts to the larger SC world at some point while there are a few options on the table, to make sure the solution ends up working well within the ecosystem. Not before you feel they're ready of course!

snappizz:
I support @jamshark70's proposal to make only the ATK class library a quark. The only downside is that the ATK code is spread across two repositories. Ultimately the decision is up to you (the ATK devs).

I'm not sure if it's a good idea to do the split last minute, so I suggest we postpone until after release.

joslloand:
Hi All, You'll see the commit I just made. This is the solution @mtmccrea has contributed. MathLib is removed as a dependency. The extension for writing that the ATK now adds is to Array, part of the core.

Agreeing with @jamshark70, @snappizz & others, moving the language elements to a Quark would be a _good idea_. We've already been thinking along these lines, in large part to support a development profile for the ATK able to move at an independent rate from sc3-plugins and the main sc3 base. ALSO, the ATK doesn't want to be in the position of breaking things(!) for other people.

In the short term, @mtmccrea contribution is the safest route. I've made the commit, so should all be good to go... and this PR can be abandoned.

snappizz:
Thanks Joseph!

bagong:
Fab, thank you!

jamshark70:
Very happy to see a reasonable solution!

PR #126, opened 2016-11-08
FM7: Implement all 32 DX7 algorithms 

kisielk:
I was bored and playing with FM7. Had the chart from the Volca FM handy, so here they are :)


snappizz:
that's great!

telephon:
thank you!

PR #136, opened 2017-02-27
plugins: SwitchDelay should not limit delayfactor

telephon:
This is debatable: but this commit suggests that the delayfactor can be:

1. negative (you may want to emphasize only odd harmonics at an octave
lower)

2. equal or very close to one (you may want to have an infinite decay
time for a while, or with very short delaytimes, you may easily need a
factor more than the current 0.9)

This is an API change.

### DIFF
--- a/source/RFWUGens/RFWUGens.cpp
+++ b/source/RFWUGens/RFWUGens.cpp
@@ -120,10 +120,6 @@ void SwitchDelay_next( SwitchDelay *unit, int inNumSamples ) {
         offset_timer = ENVLEN;
     }
 
-    // limit the delay multiplier to reasonable numbers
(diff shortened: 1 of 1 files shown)


telephon:
The change makes SwitchDelay behave more like CombN.

LFSaw:
the helpfile reads:
> delayfactor — multiplier for feedback level, affects the length of the feedback tail (default 0.7). limited slightly below 1.0 to avoid speaker damaging mistakes.

IMHO, SuperCollider's intention is to open up also unintentional possibilities (which might include speaker damage, although I am unsure this'd damage the speakers but rather a listener's ears or the amp).

So I'd go for that. 
Who's are RFW? @joshpar ? I saw you checked them in initially...

telephon:
(ping)

PR #138, opened 2017-04-04
DWGUGen: Convert iso-8859-1 character to UTF-8 in help

8c6794b6:
Convert encoding of DWGPluckedStiff help file from iso-8859-1 to UTF-8.

### DIFF
--- a/source/DWGUGens/sc/HelpSource/Classes/DWGPluckedStiff.schelp
+++ b/source/DWGUGens/sc/HelpSource/Classes/DWGPluckedStiff.schelp
@@ -6,7 +6,7 @@ categories:: UGens>PhysicalModels
 Description::
 
 Digital wave guide physical model of a plucked instrument.
-It implements inharmonicity due to string stiffness acording to J. Rauhala and V. V‰lim‰ki, Tunable dispersion filter design for piano synthesis, IEEE
+It implements inharmonicity due to string stiffness acording to J. Rauhala and V. V√§lim√§ki, Tunable dispersion filter design for piano synthesis, IEEE
 Signal Processing Letters, vol. 13, no. 5, pp. 253256,
 2006.
 


bagong:
So far we've (inconsistently) avoided unicode, especially in help... Let's postpone this, okay, there is a chance that things get better on the unicode from in the forseeable future... Even if it works for you, on Windows unicode chars can cause the SCDoc parser to crash. This is not personal, we've undone accidental commits containing stuff like unicode checkmarks in the past... Yes, there is a Japanese turorial ;), but it has to be removed in some use cases...

8c6794b6:
Thank you for quick response.

I'm not familiar with issues under Windows, so no objection for postponing.


For your interest ... I've found that the character "ä" appears in two schelp files: [SawDPW.schelp](https://github.com/supercollider/sc3-plugins/blob/master/source/MCLDUGens/sc/HelpSource/Classes/SawDPW.schelp#L13) and [DWGPluckedStiff.schelp](https://github.com/supercollider/sc3-plugins/blob/master/source/DWGUGens/sc/HelpSource/Classes/DWGPluckedStiff.schelp#L9), and the "ä"s in the two differs.

The "ä" in SawDPW help is Unicode (byte sequence: #xc3 #xa4). On the other hand, the "ä" in the DWGPluckedStiff help is ISO-8859-1 (byte sequence: #xe4). One may find the difference in rendered HTML of [SawDPW](http://doc.sccode.org/Classes/SawDPW.html) and [DWGPluckedStiff](http://doc.sccode.org/Classes/DWGPluckedStiff.html). The byte sequence data could be confirmed via viewing the `hexdump` of the schelp files.

This pull request replaces the "ä" in DWGPluckedStiff help with the one used in SawDPW help.

So far I've found, DWGPluckedStiff help file was the only one with ISO-8859-1 encoding. [Writing Help](http://doc.sccode.org/Guides/WritingHelp.html) guide seems like recommending UTF-8, so the existence of a file with ISO-8859-1 encoding seemed bit odd to me.

brianlheim:
@bagong Helpfiles can be safely _encoded_ in UTF-8, but the filenames are a different story. FYI I've already merged UTF-8 encoded characters into a help doc in the main repo with no issue. :) If the contents of a file include any codepoints beyond ASCII, those codepoints should be encoded in UTF-8 rather than Latin1 (ISO-8859-1).

bagong:
Yea, I wasn't fully concentrated, had alarm bells ring ;) Just came across your emoticons (on Mac). Nevertheless: running SCWin in debug-mode will crash the scdoc-parser if the Japanese tutorial is among the help files. That observation still holds.

PR #141, opened 2017-04-25
adding Fasor - a foldable phasor

redFrik:
previously called RedPhasor and RedPhasor2 - phasors that can loop.
i put them in PhasorUGens were we can later add more phasor variants.


source/PhasorUGens/Fasor.cpp
```diff
@@ -0,0 +1,1016 @@
+/*
+	SuperCollider real time audio synthesis system
+ Copyright (c) 2002 James McCartney. All rights reserved.
```
brianlheim: Should be updated with actual author name.

source/PhasorUGens/Fasor.cpp
```diff
@@ -0,0 +1,1016 @@
...
+    unit->m_previn= ZIN0(0);
+    unit->ppdir= 1;
+    ZOUT0(0)= unit->mLevel= ZIN0(2);
+}
+
+void Fasor_next_kk(Fasor *unit, int inNumSamples) {
+    float *out= ZOUT(0);
+    float in= ZIN0(0);
+    float rate= sc_max(0, ZIN0(1));
+    double start= ZIN0(2);
```
brianlheim: Forgive my ignorance, but shouldn't these be floats as well?

redFrik:
thanks for comments. the code is +10 years old and my memory of it is vague so sorry for that. e.g. i can't remember why all looppoints are doubles.
the length of the code is basically what it is with two ugens * taking care of all kk, ak, aa variants * 3 looping modes. yes, it is possible to rewrite and make it more compact, but time, motivation, risk of introducing bugs etc makes me reluctant. at least won't happen this week.

telephon:
The case distinctions are sometimes a bit subtle, so I understand you don't want to touch it. I'm fine with it, many ugens are explicitly unfolded like this (and I have my doubts about e.g. introducing template C++ code where it isn't absolutely necessary). 

It might nevertheless bring some clarity about the behavior if you could factor out the combinations and see if they are actually not so complicated in the end. In any case, both is fine.

LFSaw:
These are really great additions to the user-open library of UGens, thanks @redFrik !

General question: Do we want to follow the formal "request changes" policy in sc-plugins?
IMHO,  this is a user-open library of experimental UGens, which should be invite people to add their own chunks of code.
We should make sure they do not crash the system and, with naming, we should be careful as well. But otherwise, and implementation-wise, IMHO no-one should be forced to adopt a specific style here and also does not need to defend his style...

Naming is IMHO a different thing... because with names, we need to be careful :)

danstowell:
I agree 100% with the attitude proposed by @LFSaw. (But I'm not currently active, so take that for what it's worth.) It's useful to ask questions about each other's code, but mandatory changes only really for the reasons LFSaw mentions.

brianlheim:
OK. I wasn't asking for mandatory changes. If sc3-plugins is meant to be a giant blob of unmaintainable code I have no complaints; it's less for me to worry about.

danstowell:
@brianlheim OK no problem - sorry I think github's interface tends to make it look like requested changes are mandatory (!)

jamshark70:
I tend to agree with LFSaw... for good or ill, sc3-plugins are (whether de facto or by design) "use at your own risk."

Unfortunately that results in situations such as, PV_PlayBuf and PV_RecordBuf have been broken forever in Linux, but... tightening it down too much means fewer contributions, so I can live with occasional issues.

LFSaw:
@jamshark70 @danstowell thanks for chiming in, and, @brianlheim thanks for all your (surely well-meant) contributions :) 
I am quite happy to hear that we seem to be on the same level here... and yes, the github interface makes no differentiation between "hey, you might want to look at this" and "This needs to be reworked".
So I guess, we have to figure it out on the social level :)

As a sideline, no-one holds you back from fixing `PV_PlayBuf` and `PV_RecordBuf` for Linux... which U guess you already tried :)

jamshark70:
> As a sideline, no-one holds you back from fixing PV_PlayBuf and PV_RecordBuf...

I took a look at the time, but didn't quite understand it... maybe I'd fare better now. Those 2 UGens are the best for time stretching...

brianlheim:
Ah sorry, I think I should have been clearer in my initial review. This line:

> Do you have any interest in refactoring this or at least flattening it out a bit?

Was meant to sound more like

> It would be great if this were a little cleaner, but I know you might not want to spend time doing that, and that's OK.

Sorry for the misunderstanding!

vivid-synth:
> If sc3-plugins is meant to be a giant blob of unmaintainable code...

My understanding of how we think of a UGen in sc3-plugins is it's been vetted to be "not terrible," but not vetted enough to be considered stable for the main repo. My aspiration is to slowly do more vetting on UGens in the sc3-plugins, and move them into the main repo.

vivid-synth:
Maybe we can merge this PR and make @brianlheim 's (imo helpful) issues into a github issue? - So people know where the concerns are and anyone can take a stab at them.

telephon:
sounds good.

telephon:
I kind of like that name. It is a sweet spot between correctness, brevity and humor.

patrickdupuis:
Seem to me that the name should include the word "loop". LoopPhasor? LoopPoints? Some samplers call this ping pong looping.

telephon:
could be also `PhasorLoop`

redFrik:
ok, let's forget it.  i'm not going to rewrite or change coding style.  they will still be available on my h-page under their original names (i think better for everyone actually).

LFSaw:
@redFrik , pity although I understand your frustration (?). 

Not sure this is of any help but here's my take on this current situation:

As long as SC does not have an elegant way to include 3rd-party scsynth plugins outside of the giant blob that is `sc3-plugins` (no Quarks-based install for UGen's atm), plugins distributed outside the `sc3-plugins` repo are not very visible and accessible. 

So for the moment the decision of the plugin authors is to either take the discussion/change requests from fellow dev's reading (and commenting) on PRs, or have their plugins disappearing in the open www. 

I personally tend towards the former but I also handle change requests and comments on coding style in supercollider/sc-plugins much more relaxed than in e.g. supercollider/supercollider. While there I can see the urgency to keep things straight, I recall the initial idea of `sc3-plugins` to be a playground to share ideas and concepts for new plugins/implementations so that others can try them and use them in their own ways. 
For me this implies that using the content of `sc3-plugins` is an "on your own risk", including (seemingly) quirky names and interfaces and maybe also crashes. In my understanding, `sc3-plugins` is a little like a housing community with individual rooms you can visit and try out things, you can even create your own little room. But you are not supposed to "fix" things in other rooms unless the house burns down or you get the permission of the room's inhabitants (either via an explicitly given policy –README– or through personal conversation).

If there is something of more general use emerging from `sc-plugins`, and since everything hosted there has to comply with the GPL, we can always start to debate a possible derivation of that functionality which then can be added to the SC core repo. IMHO _that_ would be the time to discuss naming schemes, coding styles etc.

Did I already mention that I find it sad how this PR development turned out?

redFrik:
thanks @LFSaw for your well formulated thoughts on sc3-plugins.  i agree, although i'm not as concerned about other plugins not being found.  e.g. i'm sure people will continue to find and use these phasors whether included in sc3-plugins or not.  this pr was more to provide a service for the community and because i had gotten multiple requests.

so na, i'm not frustrated - rather i'm performing a liberating spring cleaning.  it feels good.  discussions stalled the process (as happens) but now time's up and my motivation gone.  we shouldn't leave this open forever so better to clean up and move on - making room for other things both in my head and in the communal work here.

for me the only worrying thing is to see that we can't recognize well tested, used and requested 3rd-party plugins.  instead we get tangled up in naming and useless to-unroll-functions-or-not discussions.

jamshark70:
> instead we get tangled up in naming... discussions.

It would be unfortunate if SC lost all of its sense of humor in this regard. Lighten up, we're not data mining to prevent terrorist attacks, we're just making some noise.

(I'm being flippant, but it is one of the things I like about SC: that it isn't dry and boring like Java.)

With that said, I've noticed in my own code that I have a tin ear for method naming, and by the time I realize how awful a particular method name is, I'm already committed to it. So a bit of review is not necessarily a bad thing.

telephon:
I'd like to have this UGen.

bagong:
I think @redFrik , it's your decision, you are the *author*. If you say yes, I'll check it on the different platforms.

redFrik:
thanks but no.  it's over with. done and dusted. fork deleted.
originals live on here... https://github.com/redFrik/f0plugins

danstowell:
<3

vivid-synth:
> for me the only worrying thing is to see that we can't recognize well tested, used and requested 3rd-party plugins. instead we get tangled up in naming and useless to-unroll-functions-or-not discussions.

I appreciate your frustration @redFrik , but I believe everyone agreed to defer concerns about the latter ("to unroll functions or not") to a later date (well after merge), and there was just a bit of bikeshedding about naming. We want to strike a balance, where we're relaxed enough that people want to contribute, but strict enough that we have high-quality code years from now. I think if you say "look, I don't want to discuss naming - Fasor is yours if you want it," most of us will gladly take it.

PR #146, opened 2017-06-02
added HOAUgens

florian-grond:
Hello, I have added the HOAUgens to the SC3plugins. The Plugins include Ambisonics encoding, transforming (rotation, beamforming, mirroring) and decoding up to the 5th order. Any feedback very welcome, Florian

### DIFF
--- a/source/HOAUGens/HOABeamDirac2Hoa1.cpp
+++ b/source/HOAUGens/HOABeamDirac2Hoa1.cpp
@@ -1,0 +1,1118 @@
+//----------------------------------------------------------
+// name: "HOABeamDirac2Hoa1"
+// version: "1.0"
+// author: "Pierre Lecomte"
+// license: "GPL"
+// copyright: "(c) Pierre Lecomte 2016"
+//
+// Code generated with Faust 0.9.95 (http://faust.grame.fr)
+//----------------------------------------------------------
+
+/* link with  */
+#include <math.h>
+// If other than 'faust2sc --prefix Faust' is used, sed this as well:
+#if !defined(SC_FAUST_PREFIX)
+# define SC_FAUST_PREFIX "Faust"
+#endif
+
+//-------------------------------------------------------------------
+// FAUST architecture file for SuperCollider.
+// Copyright (C) 2005-2012 Stefan Kersten.
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License as
+// published by the Free Software Foundation; either version 2 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA
+// 02111-1307 USA
+//-------------------------------------------------------------------
+
+#include <map>
+#include <string>
+#include <string.h>
+#include <SC_PlugIn.h>
+
+/************************************************************************
+    IMPORTANT NOTE : this file contains two clearly delimited sections :
+    the ARCHITECTURE section (in two parts) and the USER section. Each section
+    is governed by its own copyright and license. Please check individually
+    each section for license and copyright information.
+*************************************************************************/
+
+/*******************BEGIN ARCHITECTURE SECTION (part 1/2)****************/
+
+/************************************************************************
+    FAUST Architecture File
+    Copyright (C) 2003-2011 GRAME, Centre National de Creation Musicale
+    ---------------------------------------------------------------------
+    This Architecture section is free software; you can redistribute it
+    and/or modify it under the terms of the GNU General Public License
+    as published by the Free Software Foundation; either version 3 of
+    the License, or (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program; If not, see <http://www.gnu.org/licenses/>.
+
+    EXCEPTION : As a special exception, you may create a larger work
+    that contains this FAUST architecture section and distribute
+    that work under terms of your choice, so long as this FAUST
+    architecture section is not modified.
+
+ ************************************************************************
+ ************************************************************************/
+ 
+/******************************************************************************
+*******************************************************************************
+
+								FAUST DSP
+
+*******************************************************************************
+*******************************************************************************/
+
+#ifndef __dsp__
+#define __dsp__
+
+#ifndef FAUSTFLOAT
+#define FAUSTFLOAT float
+#endif
+
+class UI;
+struct Meta;
+
+/**
+* Signal processor definition.
+*/
+
+class dsp {
+
+    public:
+
+        dsp() {}
+        virtual ~dsp() {}
+
+        /* Return instance number of audio inputs */
+        virtual int getNumInputs() = 0;
+    
+        /* Return instance number of audio outputs */
+        virtual int getNumOutputs() = 0;
+    
+        /**
+         * Trigger the UI* parameter with instance specific calls
+         * to 'addBtton', 'addVerticalSlider'... in order to build the UI.
+         *
+         * @param ui_interface - the UI* user interface builder
+         */
+        virtual void buildUserInterface(UI* ui_interface) = 0;
+    
+        /* Returns the sample rate currently used by the instance */
+        virtual int getSampleRate() = 0;
+    
+        /** Global init, calls the following methods :
+         * - static class 'classInit' : static table initialisation
+         * - 'instanceInit' : constants and instance table initialisation
+         *
+         * @param samplingRate - the sampling rate in Herz
+         */
+        virtual void init(int samplingRate) = 0;
+    
+        /** Init instance state
+         *
+         * @param samplingRate - the sampling rate in Herz
+         */
+        virtual void instanceInit(int samplingRate) = 0;
+    
+        /** Init instance constant state
+         *
+         * @param samplingRate - the sampling rate in Herz
+         */
+        virtual void instanceConstants(int samplingRate) = 0;
+    
+        /* Init default control parameters values */
+        virtual void instanceResetUserInterface() = 0;
+    
+        /* Init instance state (delay lines...) */
+        virtual void instanceClear() = 0;
+    
+        /**  
+         * Return a clone of the instance.
+         *
+         * @return a copy of the instance on success, otherwise a null pointer.
+         */
+        virtual dsp* clone() = 0;
+    
+        /**
+         * Trigger the Meta* parameter with instance specific calls to 'declare' (key, value metadata).
+         *
+         * @param m - the Meta* meta user
+         */
+        virtual void metadata(Meta* m) = 0;
+    
+        /**
+         * DSP instance computation, to be called with sucessive in/out audio buffers.
+         *
+         * @param count - the nomber of frames to compute
+         * @param inputs - the input audio buffers as an array of non-interleaved FAUSTFLOAT samples (eiher float, doucbe or quad)
+         * @param outputs - the output audio buffers as an array of non-interleaved FAUSTFLOAT samples (eiher float, doucbe or quad)
+         *
+         */
+        virtual void compute(int count, FAUSTFLOAT** inputs, FAUSTFLOAT** outputs) = 0;
+    
+        /**
+         * DSP instance computation : alternative method to be used by subclasses.
+         *
+         * @param date_usec - the timestamp in microsec given by audio driver.
+         * @param count - the nomber of frames to compute
+         * @param inputs - the input audio buffers as an array of non-interleaved FAUSTFLOAT samples (eiher float, doucbe or quad)
+         * @param outputs - the output audio buffers as an array of non-interleaved FAUSTFLOAT samples (eiher float, doucbe or quad)
+         *
+         */
+        virtual void compute(double date_usec, int count, FAUSTFLOAT** inputs, FAUSTFLOAT** outputs) { compute(count, inputs, outputs); }
+       
+};
+
+/**
+ * Generic DSP decorator.
+ */
+
+class decorator_dsp : public dsp {
+
+    protected:
+
+        dsp* fDSP;
+
+    public:
+
+        decorator_dsp(dsp* dsp = 0):fDSP(dsp) {}
+        virtual ~decorator_dsp() { delete fDSP; }
+
+        virtual int getNumInputs() { return fDSP->getNumInputs(); }
+        virtual int getNumOutputs() { return fDSP->getNumOutputs(); }
+        virtual void buildUserInterface(UI* ui_interface) { fDSP->buildUserInterface(ui_interface); }
+        virtual int getSampleRate() { return fDSP->getSampleRate(); }
+        virtual void init(int samplingRate) { fDSP->init(samplingRate); }
+        virtual void instanceInit(int samplingRate) { fDSP->instanceInit(samplingRate); }
+        virtual void instanceConstants(int samplingRate) { fDSP->instanceConstants(samplingRate); }
+        virtual void instanceResetUserInterface() { fDSP->instanceResetUserInterface(); }
+        virtual void instanceClear() { fDSP->instanceClear(); }
+        virtual decorator_dsp* clone() { return new decorator_dsp(fDSP->clone()); }
+        virtual void metadata(Meta* m) { return fDSP->metadata(m); }
+        virtual void compute(int count, FAUSTFLOAT** inputs, FAUSTFLOAT** outputs) { fDSP->compute(count, inputs, outputs); }
+        virtual void compute(double date_usec, int count, FAUSTFLOAT** inputs, FAUSTFLOAT** outputs) { fDSP->compute(date_usec, count, inputs, outputs); }
+       
+};
+
+/**
+ * On Intel set FZ (Flush to Zero) and DAZ (Denormals Are Zero)
+ * flags to avoid costly denormals.
+ */
+
+#ifdef __SSE__
+    #include <xmmintrin.h>
+    #ifdef __SSE2__
+        #define AVOIDDENORMALS _mm_setcsr(_mm_getcsr() | 0x8040)
+    #else
+        #define AVOIDDENORMALS _mm_setcsr(_mm_getcsr() | 0x8000)
+    #endif
+#else
+    #define AVOIDDENORMALS
+#endif
+
+#endif
+/************************************************************************
+    FAUST Architecture File
+    Copyright (C) 2003-2016 GRAME, Centre National de Creation Musicale
+    ---------------------------------------------------------------------
+    This Architecture section is free software; you can redistribute it
+    and/or modify it under the terms of the GNU General Public License
+    as published by the Free Software Foundation; either version 3 of
+    the License, or (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program; If not, see <http://www.gnu.org/licenses/>.
+
+    EXCEPTION : As a special exception, you may create a larger work
+    that contains this FAUST architecture section and distribute
+    that work under terms of your choice, so long as this FAUST
+    architecture section is not modified.
+
+
+ ************************************************************************
+ ************************************************************************/
+ 
+#ifndef FAUST_UI_H
+#define FAUST_UI_H
+
+#ifndef FAUSTFLOAT
+#define FAUSTFLOAT float
+#endif
+
+/*******************************************************************************
+ * UI : Faust User Interface
+ * This abstract class contains only the method that the faust compiler can
+ * generate to describe a DSP interface.
+ ******************************************************************************/
+
+class UI
+{
+
+    public:
+
+        UI() {}
+
+        virtual ~UI() {}
+
+        // -- widget's layouts
+
+        virtual void openTabBox(const char* label) = 0;
+        virtual void openHorizontalBox(const char* label) = 0;
+        virtual void openVerticalBox(const char* label) = 0;
+        virtual void closeBox() = 0;
+
+        // -- active widgets
+
+        virtual void addButton(const char* label, FAUSTFLOAT* zone) = 0;
+        virtual void addCheckButton(const char* label, FAUSTFLOAT* zone) = 0;
+        virtual void addVerticalSlider(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step) = 0;
+        virtual void addHorizontalSlider(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step) = 0;
+        virtual void addNumEntry(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step) = 0;
+
+        // -- passive widgets
+
+        virtual void addHorizontalBargraph(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT min, FAUSTFLOAT max) = 0;
+        virtual void addVerticalBargraph(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT min, FAUSTFLOAT max) = 0;
+
+        // -- metadata declarations
+
+        virtual void declare(FAUSTFLOAT*, const char*, const char*) {}
+};
+
+//----------------------------------------------------------------
+//  Generic decorator
+//----------------------------------------------------------------
+
+class DecoratorUI : public UI
+{
+    protected:
+    
+        UI* fUI;
+
+    public:
+    
+        DecoratorUI(UI* ui = 0):fUI(ui)
+        {}
+
+        virtual ~DecoratorUI() { delete fUI; }
+
+        // -- widget's layouts
+        virtual void openTabBox(const char* label)          { fUI->openTabBox(label); }
+        virtual void openHorizontalBox(const char* label)   { fUI->openHorizontalBox(label); }
+        virtual void openVerticalBox(const char* label)     { fUI->openVerticalBox(label); }
+        virtual void closeBox()                             { fUI->closeBox(); }
+
+        // -- active widgets
+        virtual void addButton(const char* label, FAUSTFLOAT* zone)         { fUI->addButton(label, zone); }
+        virtual void addCheckButton(const char* label, FAUSTFLOAT* zone)    { fUI->addCheckButton(label, zone); }
+        virtual void addVerticalSlider(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step)
+            { fUI->addVerticalSlider(label, zone, init, min, max, step); }
+        virtual void addHorizontalSlider(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step) 	
+            { fUI->addHorizontalSlider(label, zone, init, min, max, step); }
+        virtual void addNumEntry(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step) 			
+            { fUI->addNumEntry(label, zone, init, min, max, step); }
+
+        // -- passive widgets	
+        virtual void addHorizontalBargraph(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT min, FAUSTFLOAT max) 
+            { fUI->addHorizontalBargraph(label, zone, min, max); }
+        virtual void addVerticalBargraph(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT min, FAUSTFLOAT max)
+            { fUI->addVerticalBargraph(label, zone, min, max); }
+
+        virtual void declare(FAUSTFLOAT* zone, const char* key, const char* val) { fUI->declare(zone, key, val); }
+
+};
+
+#endif
+/************************************************************************
+ ************************************************************************
+    FAUST Architecture File
+	Copyright (C) 2003-2011 GRAME, Centre National de Creation Musicale
+    ---------------------------------------------------------------------
+    This Architecture section is free software; you can redistribute it
+    and/or modify it under the terms of the GNU General Public License
+	as published by the Free Software Foundation; either version 3 of
+	the License, or (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+	along with this program; If not, see <http://www.gnu.org/licenses/>.
+
+ ************************************************************************
+ ************************************************************************/
+ 
+#ifndef __misc__
+#define __misc__
+
+#include <algorithm>
+#include <map>
+#include <string.h>
+#include <stdlib.h>
+
+/************************************************************************
+ ************************************************************************
+    FAUST Architecture File
+	Copyright (C) 2003-2011 GRAME, Centre National de Creation Musicale
+    ---------------------------------------------------------------------
+    This Architecture section is free software; you can redistribute it
+    and/or modify it under the terms of the GNU General Public License
+	as published by the Free Software Foundation; either version 3 of
+	the License, or (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+	along with this program; If not, see <http://www.gnu.org/licenses/>.
+
+ ************************************************************************
+ ************************************************************************/
+ 
+#ifndef __meta__
+#define __meta__
+
+struct Meta
+{
+    virtual void declare(const char* key, const char* value) = 0;
+    virtual ~Meta() {};
+};
+
+#endif
+
+using std::max;
+using std::min;
+
+struct XXXX_Meta : std::map<const char*, const char*>
+{
+    void declare(const char* key, const char* value) { (*this)[key]=value; }
+};
+
+struct MY_Meta : Meta, std::map<const char*, const char*>
+{
+    void declare(const char* key, const char* value) { (*this)[key]=value; }
+};
+
+inline int lsr(int x, int n)	{ return int(((unsigned int)x) >> n); }
+
+inline int int2pow2(int x)		{ int r = 0; while ((1<<r) < x) r++; return r; }
+
+inline long lopt(char* argv[], const char* name, long def)
+{
+	int	i;
+	for (i = 0; argv[i]; i++) if (!strcmp(argv[i], name)) return atoi(argv[i+1]);
+	return def;
+}
+
+inline bool isopt(char* argv[], const char* name)
+{
+	int	i;
+	for (i = 0; argv[i]; i++) if (!strcmp(argv[i], name)) return true;
+	return false;
+}
+
+inline const char* lopts(char* argv[], const char* name, const char* def)
+{
+	int	i;
+	for (i = 0; argv[i]; i++) if (!strcmp(argv[i], name)) return argv[i+1];
+	return def;
+}
+
+#endif
+
+
+using namespace std;
+
+#if defined(__GNUC__) && __GNUC__ >= 4
+# define FAUST_EXPORT __attribute__((visibility("default")))
+#else
+# define FAUST_EXPORT  SC_API_EXPORT
+#endif
+
+//----------------------------------------------------------------------------
+// Vector intrinsics
+//----------------------------------------------------------------------------
+
+
+//----------------------------------------------------------------------------
+// Metadata
+//----------------------------------------------------------------------------
+
+class MetaData : public Meta
+               , public std::map<std::string, std::string>
+{
+public:
+    void declare(const char* key, const char* value)
+    {
+        (*this)[key] = value;
+    }
+};
+
+//----------------------------------------------------------------------------
+// Control counter
+//----------------------------------------------------------------------------
+
+class ControlCounter : public UI
+{
+public:
+    ControlCounter()
+        : mNumControlInputs(0),
+          mNumControlOutputs(0)
+    { }
+
+    size_t getNumControls() const { return getNumControlInputs(); }
+    size_t getNumControlInputs() const { return mNumControlInputs; }
+    size_t getNumControlOutputs() const { return mNumControlOutputs; }
+
+    // Layout widgets
+    virtual void openTabBox(const char* label) { }
+    virtual void openHorizontalBox(const char* label) { }
+    virtual void openVerticalBox(const char* label) { }
+    virtual void closeBox() { }
+
+    // Active widgets
+    virtual void addButton(const char* label, FAUSTFLOAT* zone)
+    { addControlInput(); }
+    virtual void addCheckButton(const char* label, FAUSTFLOAT* zone)
+    { addControlInput(); }
+    virtual void addVerticalSlider(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step)
+    { addControlInput(); }
+    virtual void addHorizontalSlider(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step)
+    { addControlInput(); }
+    virtual void addNumEntry(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step)
+    { addControlInput(); }
+
+    // Passive widgets
+    virtual void addHorizontalBargraph(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT min, FAUSTFLOAT max)
+    { addControlOutput(); }
+    virtual void addVerticalBargraph(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT min, FAUSTFLOAT max)
+    { addControlOutput(); }
+
+protected:
+    void addControlInput() { mNumControlInputs++; }
+    void addControlOutput() { mNumControlOutputs++; }
+
+private:
+    size_t mNumControlInputs;
+    size_t mNumControlOutputs;
+};
+
+//----------------------------------------------------------------------------
+// UI control
+//----------------------------------------------------------------------------
+
+struct Control
+{
+    typedef void (*UpdateFunction)(Control* self, FAUSTFLOAT value);
+
+    UpdateFunction updateFunction;
+    FAUSTFLOAT* zone;
+    FAUSTFLOAT min, max;
+
+    inline void update(FAUSTFLOAT value)
+    {
+        (*updateFunction)(this, value);
+    }
+
+    static void simpleUpdate(Control* self, FAUSTFLOAT value)
+    {
+        *self->zone = value;
+    }
+    static void boundedUpdate(Control* self, FAUSTFLOAT value)
+    {
+        *self->zone = sc_clip(value, self->min, self->max);
+    }
+};
+
+//----------------------------------------------------------------------------
+// Control allocator
+//----------------------------------------------------------------------------
+
+class ControlAllocator : public UI
+{
+public:
+    ControlAllocator(Control* controls)
+        : mControls(controls)
+    { }
+
+    // Layout widgets
+    virtual void openTabBox(const char* label) { }
+    virtual void openHorizontalBox(const char* label) { }
+    virtual void openVerticalBox(const char* label) { }
+    virtual void closeBox() { }
+
+    // Active widgets
+    virtual void addButton(const char* label, FAUSTFLOAT* zone)
+    { addSimpleControl(zone); }
+    virtual void addCheckButton(const char* label, FAUSTFLOAT* zone)
+    { addSimpleControl(zone); }
+    virtual void addVerticalSlider(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step)
+    { addBoundedControl(zone, min, max, step); }
+    virtual void addHorizontalSlider(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step)
+    { addBoundedControl(zone, min, max, step); }
+    virtual void addNumEntry(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step)
+    { addBoundedControl(zone, min, max, step); }
+
+    // Passive widgets
+    virtual void addHorizontalBargraph(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT min, FAUSTFLOAT max) { }
+    virtual void addVerticalBargraph(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT min, FAUSTFLOAT max) { }
+
+private:
+    void addControl(Control::UpdateFunction updateFunction, FAUSTFLOAT* zone, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT /* step */)
+    {
+        Control* ctrl        = mControls++;
+        ctrl->updateFunction = updateFunction;
+        ctrl->zone           = zone;
+        ctrl->min            = min;
+        ctrl->max            = max;
+    }
+    void addSimpleControl(FAUSTFLOAT* zone)
+    {
+        addControl(Control::simpleUpdate, zone, 0.f, 0.f, 0.f);
+    }
+    void addBoundedControl(FAUSTFLOAT* zone, FAUSTFLOAT min, FAUSTFLOAT max, FAUSTFLOAT step)
+    {
+        addControl(Control::boundedUpdate, zone, min, max, step);
+    }
+
+private:
+    Control* mControls;
+};
+
+//----------------------------------------------------------------------------
+// FAUST generated code
+//----------------------------------------------------------------------------
+
+#ifndef FAUSTFLOAT
+#define FAUSTFLOAT float
+#endif  
+
+
+#ifndef FAUSTCLASS 
+#define FAUSTCLASS mydsp
+#endif
+
+class mydsp : public dsp {
+  private:
+	FAUSTFLOAT 	fcheckbox0;
+	double 	fConst0;
+	double 	fConst1;
+	FAUSTFLOAT 	fentry0;
+	double 	fRec0[2];
+	double 	fRec1[2];
+	FAUSTFLOAT 	fslider0;
+	double 	fRec2[2];
+	double 	fRec3[2];
+	double 	fRec4[2];
+	FAUSTFLOAT 	fslider1;
+	FAUSTFLOAT 	fslider2;
+	int fSamplingFreq;
+
+  public:
+	virtual void metadata(Meta* m) { 
+		m->declare("name", "HOABeamDirac2Hoa1");
+		m->declare("version", "1.0");
+		m->declare("author", "Pierre Lecomte");
+		m->declare("license", "GPL");
+		m->declare("copyright", "(c) Pierre Lecomte 2016");
+		m->declare("lib/lebedev.lib/name", "Lebdev grids and weights");
+		m->declare("lib/lebedev.lib/version", "1.0");
+		m->declare("lib/lebedev.lib/author", "Pierre Lecomte");
+		m->declare("lib/lebedev.lib/license", "GPL");
+		m->declare("lib/lebedev.lib/copyright", "(c) Pierre Lecomte 2014");
+		m->declare("lib/ymn.lib/name", "Spherical Harmonics library");
+		m->declare("lib/ymn.lib/version", "1.0");
+		m->declare("lib/ymn.lib/author", "Pierre Lecomte");
+		m->declare("lib/ymn.lib/license", "GPL");
+		m->declare("lib/ymn.lib/copyright", "(c) Pierre Lecomte 2016");
+		m->declare("lib/nfc.lib/name", "NF And NFC Filters Library");
+		m->declare("lib/nfc.lib/version", "1.0");
+		m->declare("lib/nfc.lib/author", "Pierre Lecomte");
+		m->declare("lib/nfc.lib/license", "GPL");
+		m->declare("lib/nfc.lib/copyright", "(c) Pierre Lecomte 2014");
+		m->declare("lib/gui.lib/name", "GUI Library");
+		m->declare("lib/gui.lib/version", "1.0");
+		m->declare("lib/gui.lib/author", "Pierre Lecomte");
+		m->declare("lib/gui.lib/license", "GPL");
+		m->declare("lib/gui.lib/copyright", "(c) Pierre Lecomte 2016");
+		m->declare("basic.lib/name", "Faust Basic Element Library");
+		m->declare("basic.lib/version", "0.0");
+		m->declare("math.lib/name", "Faust Math Library");
+		m->declare("math.lib/version", "2.0");
+		m->declare("math.lib/author", "GRAME");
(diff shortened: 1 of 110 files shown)


florian-grond:
The FAUST code has been implemented by Pierre Lecomte and is here on GitHub:
https://github.com/sekisushai/ambitools

snappizz:
could you add a README to HOAUgens that links to the repo? after that we're set

florian-grond:
done!

florian-grond:
Hi bagong, I recompiled some of the HOAUgens with the most recent FAUST version, also added HOAUGens to CMakeLists.txt where I also added:
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fbracket-depth=512")
as some of the autogenerated code for higher orders generates deep nesting.
Do I need to make a new PR?

source/HOAUGens/sc/Classes/HOAAzimuthRotator2.sc
```diff
@@ -1,4 +1,4 @@
-HOARotatorAz2 : MultiOutUGen
+FaustHOAAzimuthRotator2 : MultiOutUGen
 {
   *ar { | in1, in2, in3, in4, in5, in6, in7, in8, in9, azimuth(0.0) |
```
telephon: @reading this, I was wondering if it wouldn't be better to just pass an array? Like 

```
*ar { | inputs, azimuth(0.0) |
    if(inputs.size < 9) { Error("not enough inputs").throw };
    ^this.multiNewList('audio', inputs ++ azimuth)
}

```

And if you define `numInputChannels` in each class (see comment further below), you can make this more general:

`  if(inputs.size < this.numInputChannels) { Error("not enough inputs").throw };`

source/HOAUGens/sc/Classes/HOAAzimuthRotator1.sc
```diff
@@ -25,6 +25,6 @@ HOARotatorAz1 : MultiOutUGen
       ^this.initOutputs(4, rate)
   }
 
-  name { ^"HOARotatorAz1" }
+  name { ^"FaustHOAAzimuthRotator1" }
```
telephon: this is redundant, it's defined in `UGen` as: 
```
name {
		^this.class.name.asString;
	}
```

source/HOAUGens/sc/Classes/HOAAzimuthRotator3.sc
```diff
@@ -1,4 +1,4 @@
-HOARotatorAz3 : MultiOutUGen
+FaustHOAAzimuthRotator3 : MultiOutUGen
 {
   *ar { | in1, in2, in3, in4, in5, in6, in7, in8, in9, in10, in11, in12, in13, in14, in15, in16, azimuth(0.0) |
```
telephon: … same here and all of them, why not array?.

source/HOAUGens/sc/Classes/HOAEncEigenMike1.sc
```diff
@@ -0,0 +1,30 @@
...
+{
+  *ar { | in1, in2, in3, in4, in5, in6, in7, in8, in9, in10, in11, in12, in13, in14, in15, in16, in17, in18, in19, in20, in21, in22, in23, in24, in25, in26, in27, in28, in29, in30, in31, in32, gain(0.0) |
+      ^this.multiNew('audio', in1, in2, in3, in4, in5, in6, in7, in8, in9, in10, in11, in12, in13, in14, in15, in16, in17, in18, in19, in20, in21, in22, in23, in24, in25, in26, in27, in28, in29, in30, in31, in32, gain)
+  }
+
+  *kr { | in1, in2, in3, in4, in5, in6, in7, in8, in9, in10, in11, in12, in13, in14, in15, in16, in17, in18, in19, in20, in21, in22, in23, in24, in25, in26, in27, in28, in29, in30, in31, in32, gain(0.0) |
+      ^this.multiNew('control', in1, in2, in3, in4, in5, in6, in7, in8, in9, in10, in11, in12, in13, in14, in15, in16, in17, in18, in19, in20, in21, in22, in23, in24, in25, in26, in27, in28, in29, in30, in31, in32, gain)
+  } 
+
+  checkInputs {
```
telephon: You could make all of these UGens inherit from one `AbstractFaustHOAUGen`.
for each UGen, you can make a method like this:

*numInputChannels { ^32 }

and then the code below would be:
```
 this.class.numAudioInputs.do({|i| 
```
once and for all of them.

source/HOAUGens/sc/Classes/HOARotator1.sc
```diff
@@ -0,0 +1,30 @@
...
+{
+  *ar { | in1, in2, in3, in4, pitch(0.0), roll(0.0), yaw(0.0) |
+      ^this.multiNew('audio', in1, in2, in3, in4, pitch, roll, yaw)
+  }
+
+  *kr { | in1, in2, in3, in4, pitch(0.0), roll(0.0), yaw(0.0) |
+      ^this.multiNew('control', in1, in2, in3, in4, pitch, roll, yaw)
+  } 
+
+  checkInputs {
```
telephon: see comment above

source/HOAUGens/sc/Classes/HOA_Azimuth_Rotator1.sc
```diff
@@ -0,0 +1,30 @@
...
+      ^this.multiNew('audio', in1, in2, in3, in4, azimuth)
+  }
+
+  *kr { | in1, in2, in3, in4, azimuth(0.0) |
+      ^this.multiNew('control', in1, in2, in3, in4, azimuth)
+  } 
+
+  checkInputs {
+    if (rate == 'audio', {
+      4.do({|i|
```
telephon: … and so on …

bagong:
@florian-grond , I'll wait checking the Windows side until you've decided the rest is final, okay? Give me an okay once you've considered other reviews.

source/HOAUGens/sc/Classes/HOA_Azimuth_Rotator1.sc
florian-grond: Thanks for all these useful comments! 
The C++ and the .sc files are all autogenerated from a HOA FAUST implementation made by Pierre Lecomte: 
https://github.com/sekisushai/ambitools
so I plan not to change much on that level. But, once the plugins work, I'll publish an HOA Quark that wraps all this in PseudoUgens, with documentation and tutorials:
https://github.com/florian-grond/SC-HOA
 so that we don't need to deal with this low level:
http://musinf.univ-st-etienne.fr/lac2017/pdfs/13_C_D_141345N.pdf

Let me know if this answers most of your concerns.

snappizz:
let's not spend too much time perfecting the design here. sc3-plugins is a junkyard and it's okay for stuff to be unstable and incomplete

source/HOAUGens/sc/Classes/HOA_Azimuth_Rotator1.sc
florian-grond: and to answer more specifically the array question, this will be dealt with in the PseudoUGens, i.e. the b-format will be an array.

telephon:
@snappizz well if Florian likes, he can of course do it. I've deliberately only commented. No duty!

telephon:
… junkyard is not a very precise description. It reminds me a little bit of the [broken window theory](https://en.wikipedia.org/wiki/Broken_windows_theory) …

miguel-negrao:
> sc3-plugins is a junkyard

I was surprised by this also... I wasn't aware that sc3-plugins was *that* unmaintained.

telephon:
@florian-grond ah good! Would you like to make a separate issue out of this, where we discuss this?

snappizz:
sorry for going off topic here, but let's be honest with ourselves -- most stuff hasn't been touched in years, and there's no shortage of basic quality issues like uninitialized Ctor samples, unprotected calls to RTAlloc, and failure to interpolate control-rate arguments :/

i've personally hit my share of sc3-plugins landmines and just haven't had the motivation to fix them myself. i mostly use core these days.

LFSaw:
Hey there! First of all, this is awesome :)

some things:

+ IMHO, the faust sources for the UGens should be included in the repo, ideally in a submodule
+ @telephon, I think your suggestions are valid. However, those are FAUST-generated classes. Within the development of the JPverb UGens, we realised that it is a good idea to leave those alone and build wrapper classes around the "raw" UGens instead. This is what @florian-grond did with the Wrapper UGens in https://github.com/florian-grond/SC-HOA. Actually, I do not see a reason why those should be separated from the sc-plugins UGens. 
+ @florian-grond  For including the HOA Ugens into sc-plugins, I'd strongly recommend looking at the workflow we established for `JPverb` and `Greyhall`. You can find a detailed description on this in https://github.com/supercollider/sc3-plugins/blob/master/source/DEINDUGens/faust_src/README.md

florian-grond:
@LFSaw  good idea to include the sources. In the HOA case they are generated from a master .dsp script dynamically so there is not one .dsp file per cpp source. I believe a link to Pierre Lecomte's repository should be enough with a clear description of how to use his script.
This is why your template for JPVerb etc only applies in part.

@telephon @LFSaw  splitting the plugins from a Quark and additional resources is very much inspired by the ATK, the Quark will also have extensive Tutorial and Guides etc. I actually did not document the UGens as they are very many and autogenerated.

LFSaw:
@snappizz , if this is your feeling, then, well... that's your feeling about it.
Personally, I think we need a plugins-system that integrates into SC as does the Quarks system. So that people can provide their ideas of language extensions the way they think they want, possibly inviting others to review them, and, on the other hand, giving people the freedom to in(/ex)clude portions where they see them fit. So actually, we'd need integration of plugins into the quarks system :)

bagong:
@LFSaw , once @brianlheim 's path-handling reform is brought to the planned point that plugin binaries can be found anywhere where class-files can be found, that should be quite easy.

florian-grond:
I think bagong is still waiting for telephon to approve or disapprove

telephon:
@bagong are you waiting for me?

bagong:
@florian-grond , I think @telephon is among the "liberals" here, as am I ;) So I take your comment as a the okay I was waiting for. I am just checking if it builds on Windows, as we have a cross-platform commitment, no intention to act as gatekeeper, just using these fabulous buttons from Github ;)

Thinking a bit into the future: the option to have plugin binaries in class-folders is in reach now. So is the option to use travis and apveyor as build-engines for binaries that could be downloaded into Quarks. SC3 plugins could then be reduced to a binaries repository where we make sure that things build and are cross-platform. And maybe a few tests ;)

LFSaw:
I guess what I want to say is, please make sure that someone else (but you) is able to rebuild the plugin c++ code from the FAUST sources. It would be a shame if all the work would go to pieces because no-one remembers how to do this in 5 years time. 
Also, IMHO it would be good to explicitly include a link to the wrappers, if you don't include them here. IMHO, the raw sc classes are on the verge of being unusable without the (excellent) wrappers, so they should be mentioned, e.g. in the README file.


cheers
	Till

bagong:
@florian-grond ,  the good news, it builds now with VS2015. But I get duplicate classes errors? Weird that they only show on Windows:

`````
ERROR: duplicate Class found: 'FaustHOAAzimuthRotator1' 
C:\Users\Rainer\AppData\Local\SuperCollider\Extensions\SC3plugins\HOAUGens\Classes\HOAAzimuthRotator1.sc
C:\Users\Rainer\AppData\Local\SuperCollider\Extensions\SC3plugins\HOAUGens\Classes\HOA_Azimuth_Rotator1.sc

ERROR: duplicate Class found: 'FaustHOAAzimuthRotator2' 
C:\Users\Rainer\AppData\Local\SuperCollider\Extensions\SC3plugins\HOAUGens\Classes\HOAAzimuthRotator2.sc
C:\Users\Rainer\AppData\Local\SuperCollider\Extensions\SC3plugins\HOAUGens\Classes\HOA_Azimuth_Rotator2.sc

ERROR: duplicate Class found: 'FaustHOAAzimuthRotator3' 
C:\Users\Rainer\AppData\Local\SuperCollider\Extensions\SC3plugins\HOAUGens\Classes\HOAAzimuthRotator3.sc
C:\Users\Rainer\AppData\Local\SuperCollider\Extensions\SC3plugins\HOAUGens\Classes\HOA_Azimuth_Rotator3.sc

ERROR: duplicate Class found: 'FaustHOAAzimuthRotator4' 
C:\Users\Rainer\AppData\Local\SuperCollider\Extensions\SC3plugins\HOAUGens\Classes\HOAAzimuthRotator4.sc
C:\Users\Rainer\AppData\Local\SuperCollider\Extensions\SC3plugins\HOAUGens\Classes\HOA_Azimuth_Rotator4.sc

ERROR: duplicate Class found: 'FaustHOAAzimuthRotator5' 
C:\Users\Rainer\AppData\Local\SuperCollider\Extensions\SC3plugins\HOAUGens\Classes\HOAAzimuthRotator5.sc
C:\Users\Rainer\AppData\Local\SuperCollider\Extensions\SC3plugins\HOAUGens\Classes\HOA_Azimuth_Rotator5.sc
````

**Comment by florian-grond** [comment]
@LFSaw Agreed in all points. 
Will document carefully how to generate the cpp files and will mention the Quark in the read me etc.
I also have already compiled the the cpps for my machine and we got it to run on Linux too. This is now just to push for crossplatform including supernova etc.  
Once we know this works in principle, I hope we will be able to improve and update on a continuous basis (like adding more decoders etc.) 
The compilation of the higher order plugins (from Faust 2 cpp) just takes a lot of time right now. so this is just to jump through all justified hoops of all the precious review comments etc. which I hope also trigger the discussion on how to better integrate FAUST in SC. I think it is a very timely issue.

**Comment by florian-grond** [comment]
@bagong pls try again, I just deleted the redundant classes.

**Comment by bagong** [comment]
@florian-grond , more good news, classes fine now, and it builds both with VS2015 and MSYS2, the later includes supernova, yey!

Tbh, I don't know what `bracket-depth` is about, but it is ignored in msvc builds and it creates an error in the MSYS build. As both builds succeed, I assume the setting is not needed on Windows. So one way of avoiding the Windows problem is:

    if(NOT WIN32)
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fbracket-depth=512")
    endif()

But do you really need it, or are just cautious? If it augments the resource hunger of the build, it might be okay to avoid it all together? Might be good for low resource machines like Raspberry Pi...

**Comment by florian-grond** [comment]
@bagong  very cool to see the built come along!
Re: bracket nesting. I needed it for one UGen, that is not included yet when compiling under OSX. 
Also, I agree that for the RaspberryPi the slimmer the better. 
Let's opt for the solution you propose for now: 
if(NOT WIN32)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fbracket-depth=512")
endif()

we can later always keep it in and comment it, with additional info for its potential use.
With FAUST I easily see such deep nestings to be generated very easily.

**Comment by bagong** [comment]
> Also, I agree that for the RaspberryPi the slimmer the better.
> Let's opt for the solution you propose for now:

I particularly like in your contribution that you are planning to work on this and support it. That's more important than fulfilling sharply defined criteria at a certain point. 

With respect to bracket-depth: how about introducing it once you introduce that plugin you're talking about, and then trying to restrict it to the plugin that actually needs it? Is it necessarily a per project setting... Btw, what strikes me about bracket-depth, Google, stackoverflow and friends have no straightforward information ;) More of this type: "That's weird. gcc itself doesn't seem to have a -fbracket-depth option, and I can't find any other tool that has such an option. If it did, I'd suggest that the real problem is that nothing should have that many nested brackets. thumb1.md is a "Machine Description" file, but I see nothing odd on line 1615. 10873 should be a column number, but there are no lines nearly that long. Could your thumb1.md file be corrupted?"

**Comment by florian-grond** [comment]
Hi there, sorry for the delay, Pierre Lecomte updated the SC-HOA plugins all compiled with the most recent Faust version.  pls, integrate into SCplugins thanks!

**Comment by florian-grond** [comment]
Hi everyone, The HOA plugins should now be built with the most recent Faust compiler.  
In the CMakeList.txt in the sources we added two things:

if (NOT CMAKE_BUILD_TYPE MATCHES Debug)
    add_definitions(-DNDEBUG)
endif()

please use CMAKE with -DCMAKE_BUILD_TYPE=Release, in order to suppress the debug output from the Faust, generated by the CPP sources. 

set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fbracket-depth=2048")

This is necessary since Faust generated CPP sources can have (very) deep bracket nesting. 
Compiling on OSX otherwise gave an error.

**Comment by florian-grond** [comment]
I also added some info to the SC3plugins readme: on OSX the supernova plugis had to be explicitly in Extension/plugins otherwise supernova does not find them.
I also posted an issue plus patch about this for SC.

**Comment by patrickdupuis** [comment]
@florian-grond if you say this is good to go, then let's merge. sc3-plugins is an open collections of SC extensions. No need for too much consensus. 

@bagong please approve when you get a chance. @snappizz you do the honours of merging 😃

**Comment by LFSaw** [comment]
Is there a description on how to build the sources yet? Just wondering...

**Comment by brianlheim** [comment]
> In the CMakeList.txt in the sources we added two things:
> 
> if (NOT CMAKE_BUILD_TYPE MATCHES Debug)
> add_definitions(-DNDEBUG)
> endif()

I don't understand why this is necessary; NDEBUG is already defined with the default Release config for CMake.

File: source/CMakeLists.txt [hunk 7]
```diff
@@ -78,6 +78,19 @@ macro(BUILD_PLUGIN PLUGIN_NAME PLUGIN_SOURCES PLUGIN_LIBS PLUGIN_DIR)
...
+# This is to be able to switch off the debug messages coming from the Faust generated CPP sources
+# by setting to release mode -DCMAKE_BUILD_TYPE=Release .
+if (NOT CMAKE_BUILD_TYPE MATCHES Debug)
+    add_definitions(-DNDEBUG)
+endif()
+
+# This is necessary since Faust generated CPP sources can have deep bracket nesting.
+# Without setting -fbracket-depth to an appropriate level, cmake on OSX resulted in error.
+# Seems not to be needed under Linux, however -fbracket-depth throws an error in the MSYS build.
+if(NOT WIN32)
```
patrickdupuis: Trying to build this on Linux immidiately fails due to ``` c++: error: unrecognized command line option ‘-fbracket-depth=2048’```. This should only be set on macOS.

patrickdupuis:
I'm also getting the API mismatch error on SC 3.9.0.
```
ERROR: API version mismatch: /usr/local/share/SuperCollider/Extensions/SC3plugins/HOAUGens/Classes/HOAEncoder4.so
    This plugin is not compatible with SuperCollider >=3.9.0
    The plugin has not been loaded; please find or compile a newer version.
    Releases of sc3-plugins can be downloaded from https://github.com/supercollider/sc3-plugins/releases
    (Plugin's API version: 2. Expected: 3)
```

source/CMakeLists.txt
brianlheim: Or better yet only with clang

brianlheim:
This PR also adds binaries to the repository; please do not do that, we provide binaries as release assets.

source/CMakeLists.txt
florian-grond: Hi @patrickdupuis  I changed it to if(APPLE)

florian-grond:
The sources contain a readme that points to the faustfiles (a Gitrepo by
Pierre Lecomte)



www.grond.at

florian-grond:
encountered the same issue with Nicolas Bouillot, just changed this to
if(APPLE)



www.grond.at

florian-grond:
just removed the binaries from the repo



www.grond.at

brianlheim:
As I said in the comment above, the test should be for clang, not Apple. This is a feature of the compiler, not OS.

florian-grond:
Hi Brian, would you mind suggesting how to correctly test for the clang
compiler? Is this a clang version issue? I'm not terribly familiar with
this.



www.grond.at

florian-grond:
For the only other Faust generated plugin JPVerb from the Deind Ugens by
Till

The NDEBUG is set like this:

set_property(TARGET JPverb APPEND PROPERTY COMPILE_DEFINITIONS NDEBUG)

since the HOA Ugens include several dozens of files, I had set it as above.
I'm happy to change it for a better solution





www.grond.at

brianlheim:
> would you mind suggesting how to correctly test for the clang compiler?

Sure. The answer given here should work (test AppleClang _and_ Clang):

https://stackoverflow.com/questions/10046114/in-cmake-how-can-i-test-if-the-compiler-is-clang

I believe this is how we test in the SC repo's CMakeLists.

Although right now we only officially support building with AppleClang on macOS, gcc on Linux, and MSVC on Windows, if that expands in the future you might have people compiling with gcc on mac and clang on Linux. In that case the current approach will break; clang won't get the right flag on Linux, and cmake will try to add a nonexistent flag on Linux. Btw, in case the terminology is confusing, AppleClang is just Apple's slight port of clang. They use slightly different versioning that follows Xcode versioning (I believe 9.0 for AppleClang is something like 4.2 for clang) but I have never seen a significant difference in behavior.

> since the HOA Ugens include several dozens of files, I had set it as above. I'm happy to change > it for a better solution

Just don't set `NDEBUG`; it is set by CMake and there should be no reason to mess with that manually. Also, having a bunch of print statements under `NDEBUG` seems a bit suspect to me.. I would consider instead surrounding them with `#ifdef FAUST_HOA_DEBUG` or something similarly unique and adding it as a CMake option. Tying it to `NDEBUG` in a project as heterogeneous as this one is asking for trouble IMO.

patrickdupuis:
@brianlheim I looked at sc3plugin's `CMakeLists.txt`. Should it be written:
```cmake
if(APPLE AND CMAKE_COMPILER_IS_CLANG)
```

brianlheim:
Patrick IIRC that second variable is set earlier on though. It's not a default provided CMake variable

patrickdupuis:
True - `CMAKE_COMPILER_IS_CLANG` is not a cmake provided variable, but it *is* used throughout CMakeLists in order to signal the use of Clang:

- sc3plugins
```cmake
if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
        set(CMAKE_COMPILER_IS_CLANG 1)
endif()
```

- SuperCollider
```cmake
elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
    set(CMAKE_COMPILER_IS_CLANG 1)
endif()
```

florian-grond:
Thanks @patrickdupuis  and @brianlheim !

So I adopted Patrick's suggestion, hope this is OK.

I've set:
if(APPLE AND CMAKE_COMPILER_IS_CLANG)
compiled it and it works. 

As for having Print statements  under NDEBUG:
#if !defined(NDEBUG)
    Print("Faust: %s numControls=%d\n", name.c_str(), g_numControls);
#endif // NDEBUG

This is just what the CPP sources generated with Faust look like. 
I really side here with @LFSaw  :
We should not change these sources, but rather adapting the workflow to accommodate them. 
I'll likely add to this Plugins updating the sources regularly so I don't want to interfere with this toolchain. 
Hope this is now good to be merged :-)

brianlheim:
> if(APPLE AND CMAKE_COMPILER_IS_CLANG)

should be 

if(CMAKE_COMPILER_IS_CLANG)

florian-grond:
changed to if(CMAKE_COMPILER_IS_CLANG) and tested OK

source/CMakeLists.txt
```diff
@@ -78,6 +78,19 @@ macro(BUILD_PLUGIN PLUGIN_NAME PLUGIN_SOURCES PLUGIN_LIBS PLUGIN_DIR)
 endmacro()
 
 
+# This is to be able to switch off the debug messages coming from the Faust generated CPP sources
+# by setting to release mode -DCMAKE_BUILD_TYPE=Release .
+if (NOT CMAKE_BUILD_TYPE MATCHES Debug)
+    add_definitions(-DNDEBUG)
+endif()
```
brianlheim: this still needs to be removed

brianlheim:
> This is just what the CPP sources generated with Faust look like.

I was unclear; I meant you should change the generator. I wasn't advocating for modifying the sources after they've been generated. I would think you would want to have granular control over this even if it is auto-generated code.

florian-grond:
@brianlheim 
I would love to change the generator but this is beyond scope of what I can do.
I checked in detail and there is no way in the faust2supercollider script (which generates these sources) that allows me to remove the -DNDEBUG Print statement from the CPP files. (I have tried many times).
If we leave it in the CMakeList with the comments I've put, it should be ok to remove it again in the future, when not needed anymore.

brianlheim:
> I would love to change the generator but this is beyond scope of what I can do.
> I checked in detail and there is no way in the faust2supercollider script (which generates these sources) that allows me to remove the -DNDEBUG Print statement from the CPP files. (I have tried many times).

Maybe I'm misunderstanding -- does your organization not have control over this script?

> If we leave it in the CMakeList with the comments I've put, it should be ok to remove it again in the future, when not needed anymore.

Leave what in?

florian-grond:
no, my organization has no control over the script :-)

we could put something like this:

# This is to be able to switch off the debug messages coming from the Faust
generated CPP sources
# by setting to release mode -DCMAKE_BUILD_TYPE=Release .
# to be removed when the supercollider2faust script allows modifying
DNDEBUG Print messages  when creating the cpp sources

if (NOT CMAKE_BUILD_TYPE MATCHES Debug)
    add_definitions(-DNDEBUG)
endif()


or something similar





www.grond.at

florian-grond:
oh I meant: faust2supercollider

# This is to be able to switch off the debug messages coming from the Faust
generated CPP sources
# by setting to release mode -DCMAKE_BUILD_TYPE=Release .
# to be removed when the faust2supercollider script allows modifying
DNDEBUG Print messages  when creating the cpp sources

if (NOT CMAKE_BUILD_TYPE MATCHES Debug)
    add_definitions(-DNDEBUG)
endif()





www.grond.at

brianlheim:
> no, my organization has no control over the script :-)

Ok.

> if (NOT CMAKE_BUILD_TYPE MATCHES Debug)
> add_definitions(-DNDEBUG)
> endif()

There is no need to add this, it is redundant. This is the default behavior of CMake. This is the third time I have said this. (https://github.com/supercollider/sc3-plugins/pull/146#issuecomment-361078055, https://github.com/supercollider/sc3-plugins/pull/146#issuecomment-360944271).

brianlheim:
Who actually has control over the `faust2supercollider` bit? I can find one script by searching (https://github.com/vinterdotter/faust/blob/master/tools/faust2appls/faust2supercollider) but I don't know if that's what's being used. This process is somewhat opaque to me as I am not very involved in sc3-plugins territory.

shimpe:
For what it's worth, I think the relevant file to change the generated code would be in the architecture directory: 
https://github.com/vinterdotter/faust/blob/fb2090f27a306ebff9f251474947829894db5aac/architecture/supercollider.cpp

patrickdupuis:
This comment from @LFSaw (if I am reading it correctly) seems to confirm that the `-DNDEBUG` fix isn't necessary: https://github.com/supercollider/sc3-plugins/pull/149#issuecomment-360973022

florian-grond:
Hi everyone, I just tested it again without the

if (NOT CMAKE_BUILD_TYPE MATCHES Debug)
add_definitions(-DNDEBUG)
endif()

statement and @LFSaw <https://github.com/lfsaw> was right, it works without.

we will have to compile with the Release flag on though:
cmake -DSC_PATH=/Users/floriangrond/SuperCollider/ -DSUPERNOVA=ON
-DCMAKE_BUILD_TYPE=Release ..

otherwise, for each HOA plugin there will be a post like

Faust: supercollider.cpp: sc_api_version = 2
Faust: FaustHOAEncoder5 numControls=6

in the SC post window.

Very sorry for the noise!


Florian





www.grond.at

brianlheim:
The releases are compiled using the Release config. I believe the readme already says to build this way when compiling for non-dev usage; if it doesn't that's a serious oversight.

florian-grond:
@brianlheim  Brian maybe you can enlighten me:  

I just built the plugins as described in the readme with the modification for supernova:

cmake -DSC_PATH=/Users/floriangrond/SuperCollider/ -DSUPERNOVA=ON ..
cmake --build . --config Release --target install

The result is that the debug messages bubble up at server start, if I do, however:

cmake -DSC_PATH=/Users/floriangrond/SuperCollider/ -DSUPERNOVA=ON -DCMAKE_BUILD_TYPE=Release ..
cmake --build . --config Release --target install

the debug prints are gone 

so we should probably put into the readme this line:
cmake -DSC_PATH=/Users/floriangrond/SuperCollider/ -DSUPERNOVA=ON -DCMAKE_BUILD_TYPE=Release ..

brianlheim:
https://cmake.org/cmake/help/v3.10/variable/CMAKE_BUILD_TYPE.html

You have to specify build type in config stage if you're using a makefiles generator

patrickdupuis:
The default cmake build type for SC and for the plugins is RelWithDebInfo. It is normal that to get debug messages when built this way. Setting DCMAKE_BUILD_TYPE=Release is how you build SC and friends without any debugging info.

patrickdupuis:
In other words Florian, what to are seeing is totally normal and intended. Users should always build a Release type SC and plugins. Maybe this needs to be stressed more in the docs, but it is not something that needs to be part of this PR.

florian-grond:
If I understand Brian correctly he wants the release only to be in the
configuring stage, for this to happen we need the

if (NOT CMAKE_BUILD_TYPE MATCHES Debug)
    add_definitions(-DNDEBUG)
endif()

in the CMakeLists

If we really wanna kick it out there we need to build this way:

cmake -DSC_PATH=/Users/floriangrond/SuperCollider/ -DSUPERNOVA=ON
-DCMAKE_BUILD_TYPE=Release ..

and the release is before the config step.

I've checked again the faust2supercollider script and I tried all debug
flags (there are two)  but no matter which combination I tried it seems not
to translate into the:

faust -i -a supercollider.cpp -xml $OPTIONS "$SRCDIR/$f" -o
"$TMP/${f%.dsp}.cpp"

maybe this PR could still go ahead?








www.grond.at

jamshark70:
> The default cmake build type for SC and for the plugins is RelWithDebInfo. It is normal that to get debug messages when built this way. Setting DCMAKE_BUILD_TYPE=Release is how you build SC and friends without any debugging info.

Well, here's what *I* want, at least, and I'd venture other developers feel the same:

- If I get a crash, I want to be able to run a debugger without rebuilding SC, so I use RelWithDebInfo.

- But those Faust messages are *really pointless* and I should not have to see them every time I boot the server. (And this PR dramatically increases the number of UGens printing them.)

So you're saying that it's impossible to meet both wishes at the same time, *and* that this is a desirable outcome.

I can grudgingly accept the first of those conclusions but I don't quite agree with the second.

If this can't be changed, then why is RelWithDebInfo the default build type? It's rude to jam these messages down the throats of everybody who is building from source.

brianlheim:
@jamshark70 Required reading: https://stackoverflow.com/questions/24767450/using-cmake-how-do-you-determine-the-default-compiler-flags-per-build-type

RelWithDebInfo defines NDEBUG. A RelWithDebInfo build will not have these print messages.

> It's rude to jam these messages down the throats of everybody who is building from source.

According to Florian nobody in this thread is responsible for that.

jamshark70:
@brianlheim - Thanks for the resource, very helpful.

One contradiction: Patrick said "The default cmake build type for SC and for the plugins is RelWithDebInfo" but the stack overflow thread says "Be cautious, because the default build type is an empty string. So unless you specify a build type, none of the above applies."

In fact, stack overflow is correct. I just updated sc3-plugins to master and rebuilt in a new build directory, *without* setting CMAKE_BUILD_TYPE, and:

```
$ scsynth -u 57110
no more csLADSPA plugins
Found 380 LADSPA plugins
Faust: supercollider.cpp: sc_api_version = 3
Faust: FaustGreyholeRaw numControls=7
Faust: supercollider.cpp: sc_api_version = 3
Faust: FaustJPverbRaw numControls=11
...
```

That doesn't exactly look like debug messages are suppressed (i.e., I'm not imagining it).

Then:

```
$ cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo
$ make
$ sudo make install
$ scsynth -u 57110
no more csLADSPA plugins
Found 380 LADSPA plugins
JackDriver: client name is 'SuperCollider'
SC_AudioDriver: sample rate = 44100.000000, driver's block size = 1024
SuperCollider 3 server ready.
Zeroconf: registered service 'SuperCollider'
```

A-ha. So RelWithDebInfo *does* suppress the debug postings (Brian is correct), but you get this benefit only from setting the build type explicitly -- it is *not* by default, as previously asserted.

So I'll amend my question -- why is RelWithDebInfo *not* the default build type?

patrickdupuis:
Sorry @jamshark70 I was mistaken. Only SC has the default build type set to RelWithDebInfo. 

You are right about not seeing the debug messages if you build `RelWithDebInfo`. I've rebuilt the plugins to confirm. 

It seems like we should be defaulting to RelWithDebInfo like it's the case with SC. Stackoverflow seems to warn against setting a default build type. I don't quite understand why. Maybe someone else can clarify this?

README.md
```diff
@@ -48,8 +48,8 @@ Make a directory for the `cmake` build files:
 sc3-plugins/$ mkdir build && cd build
 sc3-plugins/build/$ cmake -DSC_PATH=/path/to/sc3source/ ..
 ```
-
 If no `SC_PATH` is provided the build system assumes the SuperCollider include files in `/usr/include/SuperCollider/`.
+Faust generated CPP sources for Ugens contain debug messages at server start and Synth launches. Set `-DCMAKE_BUILD_TYPE=Release` (see below) to suppress them.
```
patrickdupuis: There is more than one reason why someone would want to build `Release`. We can improve this README with regards to the suggested cmake configurations. This PR shouldn't concern itself with such things. Please leave this line out.

README.md
```diff
@@ -77,6 +79,8 @@ WARNING: on OSX, if you want to install into `CMAKE_INSTALL_PREFIX`, you have to
     * `sc3-plugins/build/$ cmake -DSUPERNOVA=ON ..`
 + Print all cmake options
     * `sc3-plugins/build/$ cmake -L ..`
++ Suppress debug output on server (typically from Faust generated CPP code)
+    * `-DCMAKE_BUILD_TYPE=Release ..`
```
patrickdupuis: Like above, leave this out.

README.md
```diff
@@ -61,6 +61,8 @@ Copy the `SC3plugins` folder to you Extensions folder (you find out which one th
 
 WARNING: on OSX, if you want to install into `CMAKE_INSTALL_PREFIX`, you have to specify it by disabling the `IN_PLACE_BUILD` cmake option which defaults to ON (see below).
 
+**Note:** supernova on OSX, supernova searches for plugins specifically in Extension/plugins and not like scsynth in Extension/anyOddFolder. If supernova on OSX cannot find your plugins make sure they are in the correct place. (patch for supernova server submitted for 3.9).
+
```
patrickdupuis: - Since this is has been fixed for 3.9.1, please remove the bit at the end (supercollider#3433). 
- The  entire note should be clearly marked as only relevant for SC <=3.9.0.
 - I think (not 100%) that this issue was present on MACOS and Windows.
- The wording is not ideal. I'm not the right person to improve upon it though.

source/CMakeLists.txt
```diff
@@ -78,6 +78,19 @@ macro(BUILD_PLUGIN PLUGIN_NAME PLUGIN_SOURCES PLUGIN_LIBS PLUGIN_DIR)
...
+
+# This is necessary since Faust generated CPP sources can have deep bracket nesting.
+# Without setting -fbracket-depth to an appropriate level, cmake on OSX resulted in error.
+# Seems not to be needed under Linux, however -fbracket-depth throws an error in the MSYS build.
```
patrickdupuis: You should mention that this is needed for Clang, and remove the bit about cmake on OSX and the entire 3rd line.

source/HOAUGens/README_HOAUgens.md
```diff
@@ -0,0 +1,19 @@
...
+(https://github.com/sekisushai/ambitools/tree/master/Faust/bin/ambitools2supercollider/HOA)
+
+In order to compile the Faust based CPP sources, we added to the CMakeLists.txt in the source folder
+the following lines:
+
+# This is to be able to switch off the debug messages coming from the Faust generated CPP sources
+# by setting to release mode -DCMAKE_BUILD_TYPE=Release
+if (NOT CMAKE_BUILD_TYPE MATCHES Debug)
+    add_definitions(-DNDEBUG)
+endif()
```
patrickdupuis: Please remove this.

source/HOAUGens/README_HOAUgens.md
```diff
@@ -0,0 +1,19 @@
...
+
+# This is necessary since Faust generated CPP sources can have (very) deep bracket nesting
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fbracket-depth=2048")
```
patrickdupuis: Mention this is needed for Clang.

florian-grond:
*@patrickdupuis * Thanks for the detailed change requests, I just addressed
all of them.



www.grond.at

brianlheim:
Sure, if you don't specify a build type at all you won't get any flags for any configuration except what's in cmake_cxx_flags. Sorry for the inaccuracy; this isn't even an option on multi config generators so I didn't consider it. You should always be specifying that when you configure IMO.

I strongly say no to setting a default build type. It was a mistake in the main repository and has led to a lot of confusion. If you don't specify a build config when building with a multi config toolchain you get Debug. If you set the default config to anything else for a single config generator it seems really inconsistent to me. Instead you should encourage people to be conscious of what they are building. But please, let's focus on this PR or start a discussion elsewhere for that.

But on a practical note it's bad to set this _and then_ use it for some other config logic because it just won't have an effect with a multi config generator, where CMake is generating for multiple configs at once.

florian-grond:
So I'll wait for @patrickdupuis  to approve the changes he requested. There is still one change request by @bagong, which I think is resolved now with the changes we made. 
I wanted to add to @jamshark70  that I also find the print messages in the post window pointless. As @brianlheim  has pointed out they are not from me but come with the Faust generated code.

The positive outcome of this admittedly lengthy conversation around this PR is that we pave the way for future Faust contributions to SuperCollider.

source/CMakeLists.txt
```diff
@@ -78,6 +78,21 @@ macro(BUILD_PLUGIN PLUGIN_NAME PLUGIN_SOURCES PLUGIN_LIBS PLUGIN_DIR)
 endmacro()
 
 
+# This is to be able to switch off the debug messages coming from the Faust generated CPP sources
+# when configuring the release mode.
+# Without this line we need to set -DCMAKE_BUILD_TYPE=Release
+# cmake -DSC_PATH=/your/path/to/SuperCollider/ -DCMAKE_BUILD_TYPE=Release ..
+if (NOT CMAKE_BUILD_TYPE MATCHES Debug)
+    add_definitions(-DNDEBUG)
+endif()
```
brianlheim: Remove this!

source/HOAUGens/README_HOAUgens.md
```diff
@@ -0,0 +1,19 @@
...
+by with
+[this script](https://github.com/sekisushai/ambitools/blob/master/Faust/src/ambitools2supercollider.sh)
+
+Since Faust generated CPP sources can have deep bracket nesting.
+Without setting -fbracket-depth to an appropriate level, clang will exit with error.
+This is why in CMakeLists there is:
+
+if(CMAKE_COMPILER_IS_CLANG)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fbracket-depth=2048")
+endif()
```
brianlheim: Remove this explanation. It is redundant and likely to decay. The comment above the relevant code in the CMakeLists file is enough.

source/CMakeLists.txt
```diff
@@ -78,6 +78,21 @@ macro(BUILD_PLUGIN PLUGIN_NAME PLUGIN_SOURCES PLUGIN_LIBS PLUGIN_DIR)
...
+
+
+# This is necessary since Faust generated CPP sources can have deep bracket nesting.
+# Without setting -fbracket-depth to an appropriate level, clang will exit with error.
+if(CMAKE_COMPILER_IS_CLANG)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fbracket-depth=2048")
+endif()
```
brianlheim: This should be set only for the relevant targets, not the whole project.

florian-grond:
*@brianlheim *is there a way to set the whole HOA folder as the target?



www.grond.at

florian-grond:
@patrickdupuis @brianlheim just merged Patrick's patch. are we good to go?

btw. a PR to Faust patching the debug flag related print statements in submitted.

florian-grond:
This just got merged after submitting together with @nicobou , should simplify the CMakeList for Faust contributions in the future:
https://github.com/grame-cncm/faust/pull/130

README.md
```diff
@@ -61,6 +60,8 @@ Copy the `SC3plugins` folder to you Extensions folder (you find out which one th
 
 WARNING: on OSX, if you want to install into `CMAKE_INSTALL_PREFIX`, you have to specify it by disabling the `IN_PLACE_BUILD` cmake option which defaults to ON (see below).
 
+**Note:** for SuperCollider <=3.9.0: supernova on MACOS and Windows searches for plugins in the Extension/plugins folder and not like scsynth in the Extension folder. If supernova cannot find the SC3plugins make sure they are in the correct place (i.e. Extension/plugins). 
+
```
patrickdupuis: **Note for SuperCollider < 3.9.1**  The Supernova server on macOS and Windows incorrectly searches for plugins in the `/Extensions/plugins` directory. The scsynth server correctly searches for plugins in `/Extensions`. If Supernova is unable to find the sc3-plugins, make sure that they are located in `/Extensions/plugins`.

Please update the text to the above.Thanks!

florian-grond: @patrickdupuis  done!

source/HOAUGens/README_HOAUgens.md
```diff
@@ -0,0 +1,11 @@
+# HOA SC3-plugins
+
+This Folder contains the HOA unit generator plugins based on the Faust library [Ambitools](https://github.com/sekisushai/ambitools) by Pierre Lecomte.  
+Use these plugins together with [SC-HOA](https://github.com/florian-grond/SC-HOA) quark by Florian Grond.
+
+The original [CPP sources are here]:
```
brianlheim: CPP -> C++

source/HOAUGens/README_HOAUgens.md
```diff
@@ -0,0 +1,11 @@
+# HOA SC3-plugins
+
+This Folder contains the HOA unit generator plugins based on the Faust library [Ambitools](https://github.com/sekisushai/ambitools) by Pierre Lecomte.  
+Use these plugins together with [SC-HOA](https://github.com/florian-grond/SC-HOA) quark by Florian Grond.
+
+The original [CPP sources are here]:  
+(https://github.com/sekisushai/ambitools/tree/master/Faust/bin/ambitools2supercollider/HOA)
+and can be generated from the
+[Faust code](https://github.com/sekisushai/ambitools/tree/master/Faust/src)
+by with
```
brianlheim: typo - "by with" -> "with"

florian-grond: Thank you @brianlheim @patrickdupuis @nicobou @sletz  @LFSaw  not to forget @sekisushai who wrote the original Faust code.
I really hope we will now be able to easily contribute more Faust generated SC3plugins in the future.

brianlheim:
Resolved merge conflicts - totally spurious issue in the main readme. should be good now.

florian-grond:
*@patrickdupuis* Oh seems like the set property for a whole directory does
not work that way, sorry I did not catch that one and just saw it in the
Travis built



www.grond.at

patrickdupuis:
@brianlheim has already identified the fix. I'm submitting a PR to your branch as I write this.

brianlheim:
Oh, I see now. Each source file in the directory gets its own target. That is problematic, and I think demonstrates how brittle the current CMake setup is for integrating plugins.

brianlheim:
not especially proud of this, but it does build:

```diff
diff --git a/source/CMakeLists.txt b/source/CMakeLists.txt
index 600210f..70fa29d 100644
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -165,6 +165,13 @@ else()
     set(INSTALL_DESTINATION_DISTRO "SC3plugins")
 endif()

+# HOAUGens
+if (CMAKE_COMPILER_IS_CLANG)
+    file(GLOB HOAUGensSources HOAUGens/*cpp)
+    foreach(HOAUGensSourceFile ${HOAUGensSources})
+        set_source_files_properties("${HOAUGensSourceFile}" PROPERTIES COMPILE_FLAGS "-fbracket-depth=2048")
+    endforeach(HOAUGensSourceFile)
+endif()

 foreach(DIR ${PLUGIN_DIRS})
     FIND_AND_BUILD_PLUGINS(${DIR})
@@ -338,11 +345,6 @@ if (NOVA_DISK_IO)
     add_subdirectory(NovaDiskIO)
 endif()

-#HOAUGens
-if (CMAKE_COMPILER_IS_CLANG)
-    set_property(TARGET HOAUGens APPEND PROPERTY COMPILE_FLAGS "-fbracket-depth=2048")
-endif()
-
 #--------
 # INSTALL
 #--------
```

sletz:
@florian-grond is this normal that the` #if !defined(NDEBUG)` are still in the cpp code? Shouldn't all Faust generated cpp files be updated with the latest https://github.com/grame-cncm/faust/commit/2affd03d5fa2d6f4b687dd7c55ba437c53bd6b23 ?

florian-grond:
@brianlheim  thanks for fixing the folder-wide settings! much appreciated!
@sletz  @nicobou and I had submitted the Faust patch on Nicolas' computer and I did not have a chance yet to rebuilt the ambitools from @sekisushai (which takes a while) with the new fasut version on my end, will do very soon.

florian-grond:
hi everyone, the most recent version just passed Travis. 
The Ugens now throw no debug messages no more and all the Faust prefixes are removed. 
This PR is still flagged as waiting for consensus. 
Hope there is no obstacle left to include it in the SC3plugins repo.

patrickdupuis:
I'll give this a test in the coming days. We will try and get this in soon :)

patrickdupuis:
I have not been able to build this PR on Rpi (`cmake -L -DCMAKE_BUILD_TYPE="Release" -DSUPERNOVA=OFF -DNATIVE=ON -DSC_PATH=../../supercollider/ ..`). Make hangs at 40%. I'm going to try again, giving the Rpi as much time as possible. Maybe things a just really slow?
```
Scanning dependencies of target HOADecBinaural3
[ 40%] Building CXX object source/CMakeFiles/HOADecBinaural3.dir/HOAUGens/HOADecBinaural3.cpp.o

source/CMakeFiles/HOADecBinaural3.dir/build.make:62: recipe for target 'source/CMakeFiles/HOADecBinaural3.dir/HOAUGens/HOADecBinaural3.cpp.o' failed
make[2]: *** [source/CMakeFiles/HOADecBinaural3.dir/HOAUGens/HOADecBinaural3.cpp.o] Interrupt
CMakeFiles/Makefile2:2683: recipe for target 'source/CMakeFiles/HOADecBinaural3.dir/all' failed
make[1]: *** [source/CMakeFiles/HOADecBinaural3.dir/all] Interrupt
Makefile:149: recipe for target 'all' failed
make: *** [all] Interrupt
```

patrickdupuis:
After several attempts, I have come to the conclusion that it is not possible to build these plugins on Rpi. I think the issue is related to a lack of memory. Perhaps someone with more experience building on embedded linux could give it a try? @redFrik perhaps?

redFrik:
i can give it a go but if it's ram related here are some things you can try...
* open `sudo raspi-config` and under advanced set memory split to 16
* build on a raspbian lite system instead of full (desktop)
* use a swap file - will be very slow but then it should really work
  * `sudo dd if=/dev/zero of=/swapfile bs=1MB count=512`
  * `sudo mkswap /swapfile`
  * `sudo swapon /swapfile`
* or as an alternative distcc

florian-grond:
I'm not surprised that this target takes a very long time to compile:

Scanning dependencies of target HOADecBinaural3

These are the ones with very deep bracket nesting, they are also heavy
on the CPU as Ugens,
if you feel like we should exclude them from the RPi built I would
have no problem with it.







www.grond.at

patrickdupuis:
@redFrik maybe this would be a better way of getting more swap memory for Rpi? https://raspberrypi.stackexchange.com/questions/70/how-to-set-up-swap-space#1605

I didn't mention it, but I am using Rasbian Lite with 16 memory split like you suggest in the build guide.

brianlheim:
This doesn't build with VS 2017:

```
"C:\Users\brian heim\code\supercollider\sc3-plugins\build\ALL_BUILD.vcxproj" (default target) (1) ->
"C:\Users\brian heim\code\supercollider\sc3-plugins\build\source\HOADecBinaural3.vcxproj" (default target) (82) ->
(ClCompile target) ->
  C:\Users\brian heim\code\supercollider\sc3-plugins\source\HOAUGens\HOADecBinaural3.cpp(1194): fatal error C1026: pars
er stack overflow, program too complex [C:\Users\brian heim\code\supercollider\sc3-plugins\build\source\HOADecBinaural3
.vcxproj]


"C:\Users\brian heim\code\supercollider\sc3-plugins\build\ALL_BUILD.vcxproj" (default target) (1) ->
"C:\Users\brian heim\code\supercollider\sc3-plugins\build\source\HOADecBinaural4.vcxproj" (default target) (83) ->
  C:\Users\brian heim\code\supercollider\sc3-plugins\source\HOAUGens\HOADecBinaural4.cpp(1412): fatal error C1026: pars
er stack overflow, program too complex [C:\Users\brian heim\code\supercollider\sc3-plugins\build\source\HOADecBinaural4
.vcxproj]


"C:\Users\brian heim\code\supercollider\sc3-plugins\build\ALL_BUILD.vcxproj" (default target) (1) ->
"C:\Users\brian heim\code\supercollider\sc3-plugins\build\source\HOADecBinaural5.vcxproj" (default target) (84) ->
  C:\Users\brian heim\code\supercollider\sc3-plugins\source\HOAUGens\HOADecBinaural5.cpp(1678): fatal error C1026: pars
er stack overflow, program too complex [C:\Users\brian heim\code\supercollider\sc3-plugins\build\source\HOADecBinaural5
.vcxproj]
```

These are the only 3 that don't build. There is no option to increase the MSVC parser stack size that I've found.

I'm not sure how you'd like to handle this. It would be very nice if the code generator didn't create expressions so complex. Even just splitting up expressions that are several thousand subexpressions long like:

image

would help.

redFrik:
for me on a rpi3 with 1gb ram it stopped building at...
[ 66%] Building CXX object source/CMakeFiles/HOADecBinaural5.dir/HOAUGens/HOADecBinaural5.cpp.o
"virtual memory exhausted: Cannot allocate memory"

then i ran make again and i got...
c++: internal compiler error: Killed (program cc1plus)

even with a 1024mb swapfile i couldn't get it to compile.

so yes, a build option flag for skipping these plugins is essential else we cannot build sc3-plugins on small systems.

sletz:
"This doesn't build with VS 2017:" : the Faust compiler can indeed generated very complex code, and this is not so easy (for now at least...) to adapt the code generation strategy in a automatic manner. 
@florian-grond  can you point me again to the original Faust DSP files, and explain how the C++ files are produced ? HOADecBinaural3 and so on..)

sekisushai:
@sletz The HOADecBinauralXX involve massive linear convolution, which explain the difficulty of compilation, especially at higher orders. 
The .dsp file is [here](https://github.com/sekisushai/ambitools/blob/master/Faust/src/hoa_decoder_lebedev50_binaural.dsp).
The .cpp files are generated with `faust2supercollider` script with `-ks` flag

florian-grond:
Let's see if @sletz  can propose something otherwise we could simply exclude the files for small systems as proposed by @redFrik 
HOADecBinaural3.cpp <- maybe exclude
HOADecBinaural4.cpp <- definitely exclude
HOADecBinaural5.cpp <- definitely exclude

The Binaural Ugens op to order 5 are certainly fun to have for powerful computers but will very likely not be used on the RPi as they are also very CPU demanding as Ugens.

sletz:
I cannot not even produce the HOADecBinaural5.cpp here.... So yes probably better to exclude them for now.

snappizz:
hey all, i must admit i've completely lost track of the current status of this PR. can anyone summarize what's up?

joshpar:
Anyone have an update on this? Do we still want it merged into the main repo, or are there concerns that still need to be resolved?

florian-grond:
Thanks for checking in, the biggest concern was that it contains a Ugen
that does not compile on the Raspberry. And I did not find the time to work
on it. Also, there have been some improvements in the underlying Faust code
recently, and there are more improvements to come. The Faust code will
reach version 1.0, in the next months according to Pierre Lecomte. I
propose to wait until then.

Florian



www.grond.at

djiamnot:
I have observed one issue with HOAUgens plugins. It breaks Class Library compilation due to a missing dependency (MathLib). I am not quite familiar with the sc3 plugins architecture and pipeline but it seems that there needs to be some dependency resolution. The compilation breaks on .sc classes as they use Cartesian() method from the aforementioned lib. Installing MathLib prior to merging `florian-grond:master`, works fine.

patrickdupuis:
I tested this out yesterday and I couldn't reproduce. I see no use of MathLib in any of the classes included in this PR. The class compilation issues you're having are probably related to something else. Try and test on a clean system.
________________________________

patrickdupuis:
Just a heads up, you might get duplicate class errors due to the fact that some files in this PR were renamed (slightly) at some point (the HOABeamDirac2... ones).

patrickdupuis:
I've created a PR against @florian-grond's branch to help move this PR forward. florian-grond/sc3-pluginsHOA#5

I've copied my PR's description below:

> 
>
>This PR adds an option that allows us to conditional build the HOAUGens. The reasons for this option being set to `OFF` by default are the following:
>
>- These plugins cannot be build on Windows MSVC due too some files having very deep bracket nesting.
>- Embedded Linux systems like Rpi aren't able to build, likely due to their limited resources.
>- As is, the sc3-plugins are build-able on all systems. Having this option set to `ON` by default would change that. It's better to require users to opt-in to these plugins.
>
>This PR is a starting point. I'm hoping to bring renewed attention to this supercollider/sc3-pluings#146, and get things merged soon.

djiamnot:
Thanks @patrickdupuis for clarification. I was, in fact, using a "contaminated" copy of the sc3-plugins.

patrickdupuis:
Whoa! I didn't see that coming :smile:

telephon:
was it ok? I am sorry, if that was too quick. I had assumed all was actually done.

florian-grond:
Excellent! very happy, Thanks Patrick for finding a way to make it
compatible!  Florian



www.grond.at

patrickdupuis:
@telephon yes, it is fine. I had just added a README to the PR before you merged.

LFSaw:
so, default flag for this is "Off"?

telephon:
Because the temperature is rising, I was worried that it would get stale before we could eat it ...

telephon:
>so, default flag for this is "Off"?

as it seems, yes. Should it perhaps be `ON`?

LFSaw:
no, leave it Off please. For now at least (since it does not compile on all platforms)

telephon:
yes, that was my thinking.

florian-grond:
I just made some suggestions offlist how we might be able to partially
switch it on, let's keep it off for now




www.grond.at

On Mon, Jun 18, 2018 at 10:01 PM, Julian Rohrhuber <notifications@github.com

telephon:
actually, trying to switch it on in my usual build folder seems not to work. Maybe I'm doing something wrong.

`cmake -DSC_PATH=/Volumes/xyz/supercollider -HOA_UGENS=ON ..`

and then 

`cmake -L`

shows me

`HOA_UGENS:BOOL=OFF`

patrickdupuis:
try with a second build folder called `build2` and see if you get the same result.

telephon:
still the same

patrickdupuis:
Oh! it's `-DHOA_UGENS` (you forgot the D)

brianlheim:
> The HOAUGens are unable to build using MSVC due a limitation in the depth of bracket nesting that is permitted by this compiler.

Almost all the plugins _do_ build with MSVC, it's only three that have a problem. Is there any plan to provide a convenient way to skip the bad ones (HOADecBinaural3/4/5)? Such as just excluding those plugins when MSVC is detected? How do you expect anyone to provide a testable build otherwise? Same with RPi, sounds above like HOADecBinaural5 doesn't build there and the rest could be used.

patrickdupuis:
Yes - this is what @florian-grond is thinking about doing as a next step. For now, though, this is the way it had to be in order to move this PR forward. Maybe we should open an issue?

brianlheim:
> For now, though, this is the way it had to be in order to move this PR forward.

??? Is there a link to a discussion? There is no consensus in the discussion above.

jamshark70:
At least the PR appears to be incomplete. AFAICS the readme does not explain the new cmake option nor that HOAUGens in their current state should not be expected to build in Visual C or perhaps on RPi as well.

Without this documentation as a bare minimum (or full compatibility), merging was premature.

patrickdupuis:
I agree that this shouldn't have been merged as is. I was intending to discuss this PR at the next dev meeting and reaching a consensus.

At this point, we should be suggesting necessary changes (maybe in #197) so we can make them happen.

telephon:
Sorry, my misunderstanding. I had understood that the cmake option is off by default, and the changes in themselves were very well maintained with a clear responsibility.

So as far as I can tell, what is really missing is a few sentences of documentation of this option and its limitations.

florian-grond:
As Brian has pointed out, almost everything compiles except for a few Ugens
HOADecBinaural3/4/5,
1) they either cause a problem on Windows because of the deep bracket
nesting
2) they cause a problem on small platforms like the RPi because compilation
is resource intensive.

These two problems will likely come back as new plugins are being developed
in HOA.
Not for many but for some plugins.
At some point in the future, the Faust compiler might not create these big
expressions anymore, but we should not wait for that.

On platforms like the RPi, and other small boards, it is very unlikely that
one would do Ambisonics work with very high orders.
I believe it would be possible to detect MSVC but to detect the RPi seems
more tricky.

Things like HOADecBinaural3/4/5 are useful but not absolutely necessary.

I would go for Brian's suggestion:
Let's explicitly exclude certain orders of UGens from the official built
with an OFF flag so that it runs on all supported platforms but for those
who compile the plugins everything is in the public repo and can be
included if desired.

for now:
if (X in HOADecBinauralX.ccp > 2)  -> do not compile
or more general:
if (X in HOAWhateverX.ccp > 2)  -> do not compile

If anyone could suggest a way to do that correctly in the CMakeList file
that would be great.


Florian










www.grond.at

patrickdupuis:
@florian-grond let's leave this thread and move the discussion to #197.

PR #148, opened 2017-06-16
Topic/atk quarkify

joslloand:
The ATK team is seeking to Quark-ify the language side components of the ATK, briefly described [here](https://github.com/ambisonictoolkit/atk-sc3/issues/56).

Class, Extensions and Help are moved to the [atk-sc3 Quark](https://github.com/ambisonictoolkit/atk-sc3), while UGens remain with sc3-plugins. Most ATK development is currently happening on the language side, rather than with the UGen plugins, so this change allows us to move development ahead quicker.


telephon:
Great! This will make it easier to grow. 

If I see this correctly, you are also removing the help files for the UGens which remain in sc3-plugins - are you sure this is intended?

joslloand:
@telephon wrote:

> If I see this correctly, you are also removing the help files for the UGens which remain in sc3-plugins - are you sure this is intended?

I'll let @mtmccrea add a response here to confirm.

The idea is that by removing the Help and Class files, the UGens don't appear to the user until the [atk-sc3](https://github.com/ambisonictoolkit/atk-sc3) Quark is installed. Also, keeping all the Help and Class files in the Quark means that the documentation and language components all are managed in one place.

I'll add @scztt as having been in on this discussion.

mtmccrea:
Yes - with help files in the quark they will be updated and distributed much more easily and in fact are the files updated most frequently. In any case the quark will be necessary to use the ATK at all because the UGen classes and other ATK classes currently live in the same file. Thanks all for having a look!

telephon:
I see, this makes sense. Because this is a little unusual, please leave a readme  in the sc3-plugins subfolder, with a note where to find the class files.

muellmusik:
As an afterthought, it just occurs to me that this highlights the continuing awkwardness of UGen quarks. But Quarks can now support different downloads for different platforms, yes? So wouldn't a better solution be to keep them as a single Quark and provide binaries for Mac and Windows, and source for Linux, or those who want to build themselves? Even better would be to automate compilation, i.e. basically work like most package managers.

telephon:
You mean to add the binaries to the quark itself, in different versions?

muellmusik:
Yes. So if you downloaded Quark ver. X it would just get the correct version of binaries for your platform. This is after all a solved problem, and that's the generally preferred solution! ;-)

telephon:
I thought that binaries are somehow "not recommended" on github?

bagong:
I think a good way would be to keep sc3-plugins as a) a repository of vetted binaries (even if opinions as to quality standards vary ;) ) that at least kind of work and do not contain viruses and b) provides automated builds for mac, linux and osx. If we'd leave the build process the responsibility of the plugins owner, we'll likely get a lot of feedback with broken, incomplete builds.

I think four things would be required to get this:

- complete build scripts for mac, osx and windows for travis and appveyor
- breaking up the current single release archive into as many archives as there are plugins packages
- add a mechanism to quarks that downloads and extracts the correct archive from sc3-plugins
- ideally the capability of scsynth and supernova to find plugins in any folder where a quark might be

One could then decide whether or not to allow "official quarks" to contain plugins from non official sources. I think security advisors would strongly recommend not to. And I am pretty sure that not many people can provide binaries for 3 platforms.

muellmusik:
> On 19 Jun 2017, at 13:55, Julian Rohrhuber <notifications@github.com> wrote:
> 
> I thought that binaries are somehow "not recommended" on github?
> 
> 
> 
Not in Git, but it does support binary releases.

joslloand:
@telephon 

Agreed, this is an _unusual_ solution. We have left the README in the /sc folder. Included is a section on [installation](https://github.com/supercollider/sc3-plugins/blob/master/source/ATK/sc/README.md#installing).

mtmccrea:
@muellmusik @bagong @scztt 

A while back, I did try auto-building the ATK plugins in a standalone repo as a [messy proof of concept](https://github.com/mtmccrea/atk-plugins-build-test). Using travis and the build scripts from SC3Plugins as a submodule (so it could still be centrally maintained), and I got it to crank out osx and linux [releases](https://github.com/mtmccrea/atk-plugins-build-test/releases) of just the ATK plugins.  It was a big hack but proved possible. 

I didn't see a viable way it could be adopted unless others with more knowledge of the needs of different platforms and more involved plugins with library/build dependencies could do it the "right way".

I also did a successful test of the Quark system pulling down pre-built binaries with the classes, etc., though SC couldn't see the binaries in the downloaded-quarks directory. That would seem to require small change to where SC looks for plugins, but as discussed above, binaries probably aren't  recommended to host in github.  

Though if the quark install process could unpack binary releases from a quark's repo, this could be combined with the above method of auto-building plugins within a quark repo. Classes, source and plugins could live in the same place, and with some careful handling of a central plugin-building submodule, folks are left to develop and update their full package and release whenever...

Lots of details and trials to get from here to there though!  ;)

brianlheim:
@mtmccrea I haven't read the rest of this convo, but just wanted to jump in and say I am hoping to modify scsynth soon so that it searches for plugins in all directories included language-side, which would include quarks folders. It may not happen for a few months, but I'm hoping to get it done by the end of the summer.

mtmccrea:
@brianlheim 
Very cool... a step in the right direction for this!

Issue #151, opened 2017-07-25
FM7: help file doesn't specify "algo" must be fixed


Arseniusz:
```SuperCollider
    SynthDef.new(\fm_synth, {
        |freq, alg = 0|
        var sig, freq_matrix, envelope;
        freq_matrix = [
            // freq, phase, amp
            [freq, 0, 1],
            [freq * 2, 0, 0.5],
            [freq * 4, 0, 0.25],
            [5009, 0, 0.1],
            [20, 0, 1],
            [30, 0, 1]
        ];
        envelope = EnvGen.kr(Env.perc(level: 0.03), doneAction: 2);
        sig = FM7.arAlgo(alg, freq_matrix, 1.5) * envelope;
        Out.ar(0, sig);
    }).add;
```

If I trying to send 'algo' argument in FM7.arAlgo I got an error:

ERROR: Primitive '_BasicAt' failed.
Index not an Integer
RECEIVER:
Instance of Array {    (0x55f9a39fdb38, gc=F4, fmt=01, flg=00, set=05)
  indexed slots [32]
      0 : instance of Function (0x55f9a424e258, size=2, set=2)
      1 : instance of Function (0x55f9a4616938, size=2, set=2)
      2 : instance of Function (0x7fbe280813f8, size=2, set=2)
      3 : instance of Function (0x55f9a4a82a78, size=2, set=2)
      4 : instance of Function (0x55f9a4619aa8, size=2, set=2)
      5 : instance of Function (0x55f9a49dbea8, size=2, set=2)
      6 : instance of Function (0x55f9a34bccc8, size=2, set=2)
      7 : instance of Function (0x55f9a4656808, size=2, set=2)
      8 : instance of Function (0x55f9a44168a8, size=2, set=2)
      9 : instance of Function (0x55f9a4a0acb8, size=2, set=2)
     10 : instance of Function (0x55f9a4a0e988, size=2, set=2)
     11 : instance of Function (0x55f9a4a16ef8, size=2, set=2)
     12 : instance of Function (0x55f9a48c33a8, size=2, set=2)
     13 : instance of Function (0x55f9a49029a8, size=2, set=2)
     14 : instance of Function (0x55f9a468fcf8, size=2, set=2)
     15 : instance of Function (0x55f9a4a05378, size=2, set=2)
     16 : instance of Function (0x55f9a46500c8, size=2, set=2)
     17 : instance of Function (0x55f9a468d678, size=2, set=2)
[...]

PR #152, opened 2017-07-26
Update description for algo argument in arAlgo method

Arseniusz:
Fix for issue - https://github.com/supercollider/sc3-plugins/issues/151


telephon:
I think it would be relatively easy to make this modulatable. Whenever I have the time, I'll take a look.

PR #156, opened 2017-08-11
RMS: fix instantiation error by calculating initial sample in ctor

LFSaw:


LFSaw:
@telephon , would you mind to skim over the change?

LFSaw:
since no-one objects (and cares), I merge.

PR #165, opened 2017-10-24
Make README clearer

brianlheim:
See https://github.com/supercollider/supercollider/issues/3250

- Remove `cwd/sc3-plugins$`-style text in shell commands (confuses some users)
- Add a troubleshooting note about missing files during build (common mistake)
- Rearrange some text to improve flow
- Fix minor typos

@unculcated, what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,85 +6,94 @@ An installation extends the functionality of SuperCollider by additional UGens t
 
 Please note that the UGens in this repository are, on average, less stable and well-maintained than the core collection. Use at your own risk!
 
-**Note:** Extensions for the SuperCollider programming language are something structurally different. They are collected within the [Quarks](http://quarks.sourceforge.net/) packaging system.
+**Note:** Extensions for the SuperCollider programming language are different. They are collected within the [Quarks](https://github.com/supercollider-quarks/) packaging system.
 
 To learn how to write your own plugins, see [example-plugins](https://github.com/supercollider/example-plugins) and the "[Writing UGens](http://doc.sccode.org/Guides/WritingUGens.html)" helpfile.
 
 For community discussion and support see the [SuperCollider mailing lists](http://www.birmingham.ac.uk/facilities/BEAST/research/supercollider/mailinglist.aspx) and the [github issue tracker](https://github.com/supercollider/sc3-plugins/issues).
 
-## Download
+## Release Downloads
 
-Compiled releases are available from the [github release page](https://github.com/supercollider/sc3-plugins/releases). For older versions (2013). see the [sourceforge project page](http://sourceforge.net/projects/sc3-plugins/files/).
-
-To compile from source (see below), either download a tarball or clone this repository:
-
-```shell
-$ git clone --recursive https://github.com/supercollider/sc3-plugins.git
-```
+Compiled releases are available from the [GitHub release page](https://github.com/supercollider/sc3-plugins/releases). For older versions (2013), see the [Sourceforge project page](http://sourceforge.net/projects/sc3-plugins/files/).
 
 ## Binary install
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,93 +6,101 @@ An installation extends the functionality of SuperCollider by additional UGens t
 
 Please note that the UGens in this repository are, on average, less stable and well-maintained than the core collection. Use at your own risk!
 
-**Note:** Extensions for the SuperCollider programming language are something structurally different. They are collected within the [Quarks](http://quarks.sourceforge.net/) packaging system.
+**Note:** Extensions for the SuperCollider programming language are different. They are collected within the [Quarks](http://quarks.sourceforge.net/) packaging system.
```
LFSaw: Maybe

> Extensions for the SuperCollider programming language are different than the server plugins collected in this repository.

?

README.md
```diff
@@ -6,93 +6,101 @@ An installation extends the functionality of SuperCollider by additional UGens t
...
+git clone --recursive https://github.com/supercollider/supercollider.git
 ```
 
-If no `SC_PATH` is provided the build system assumes the SuperCollider include files in `/usr/include/SuperCollider/`.
+Be sure to use `--recursive`, or to initialize the submodules after cloning, otherwise you will not have files
+that are necessary to build the project.
+
+Like SuperCollider, sc3-plugins uses `cmake`. Execute the following commands in the directory where you cloned sc3-plugins;
+replace `/path/to/sc/` with the path to the SuperCollider source directory. If you cloned both repositories from the same
+location, that will probably be `../supercollider`.
```
LFSaw: > from the same working directory

(location is misleading, sounds like the git uri)

README.md
```diff
@@ -6,93 +6,101 @@ An installation extends the functionality of SuperCollider by additional UGens t
...
+make
+make install
 ```
 
-On OSX, the plugins will end up in `sc3-plugins/build/SC3plugins`.
-Copy the `SC3plugins` folder to you Extensions folder (you find out which one that is by evaluating `Platform.userExtensionDir` from within SuperCollider).
+If no `SC_PATH` is provided the build system assumes the SuperCollider include files are in `/usr/include/SuperCollider/`.
+
+On macOS, the plugins will end up in `sc3-plugins/build/SC3plugins`.
+Copy the `SC3plugins` folder to your Extensions folder (evaluate `Platform.userExtensionDir` in SuperCollider to find it).
```
LFSaw: Link (`ln -s`) is a better idea actually...

README.md
```diff
@@ -6,93 +6,101 @@ An installation extends the functionality of SuperCollider by additional UGens t
...
+
+#### Starting over
 
 If something went wrong and you want to start from scratch, delete everything in the build directory that you made:
 
 ```shell
-sc3-plugins/build/$ make uninstall # only if you did `make install` before
-sc3-plugins/build/$ rm -r *
+make uninstall # only if you ran `make install` before
+rm -r *
```
LFSaw: super dangerous!
how about 

```
cd ..
rm -r build
```
?

README.md
snappizz: there's also a link here that needs to be updated

LFSaw: https://github.com/supercollider-quarks/

README.md
brianlheim: i just don't think i agree in this case. user might very well delete the source directory after building.

brianlheim:
Thanks for the suggestions @LFSaw - changes have been made

brianlheim:
ping. can someone please check this? if we announce a beta lots of people are going to be looking at the readme

README.md
LFSaw: very well

PR #168, opened 2017-10-29
Updates to docs and packaging

### LINKED ISSUE #166: README for compiled plugins
Hello,

The compiled plugins for apple does not include any kind of documentation. A simple README file with installation instructions would be a welcome addition to that release.  The current instructions are hard to locate from the release page and are part of a LOT of text about compiling and other issues, which is somewhat difficult for new users.

(Ideally, plugins could be integrated into the quarks system somehow, but I'm aware this is not going to happen soon.)

brianlheim:
- Add info about packaging/release process in a new DEVELOPING.md
- Add a configurable README for MacOS which can be included in releases
- As a result, Fixes #166
- Make a slight tweak to the readme that was unnecessarily Linuxy

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,10 +135,22 @@ if (OSX_PACKAGE)
   add_subdirectory(osx_package)
 endif()
 
-if(WIN32)
-    configure_file( README_WINDOWS.txt.in README_WINDOWS.txt)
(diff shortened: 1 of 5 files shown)


brianlheim:
cc @celesteh @LFSaw @snappizz

PR #175, opened 2018-01-14
Compatibility with BSD systems

yurivict:
* CMAKE_SYSTEM_NAME match
* Use of ${CMAKE_DL_LIBS}

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -146,7 +146,7 @@ endif()
 GET_GCC_VERSION(gcc_version)
 
 
-if (CMAKE_SYSTEM_NAME MATCHES "Linux")
+if (CMAKE_SYSTEM_NAME MATCHES "Linux|.*BSD|DragonFly")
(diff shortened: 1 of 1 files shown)


yurivict:
You're welcome!

(Upstreaming the patch from the FreeBSD port.)

patrickdupuis:
@yurivict looks like cmake is failing. @brianlheim suggested over on slack that `“${CMAKE_DL_LIBS}”` (writen in quotes) may be the solution.

yurivict:
>  looks like cmake is failing

Hmm, it doesn't fail for me.
You can also see examples of use in other projects in the same way: 
https://github.com/awesomeWM/awesome/issues/82
https://gitlab.maisondelasimulation.fr/l2c/l2c/commit/b2f3e12c7050f19a3cf591e8cfe925b67115bfbe

brianlheim:
> Hmm, it doesn't fail for me.

That's great but we're talking about our CI

brianlheim:
Fantastic, thank you!

PR #177, opened 2018-01-20
readme: make installation instructions friendlier

### LINKED ISSUE #176: Install process difficult to follow
I run quite a lot of workshops etc dependent on supercollider with sc3-plugins, and have observed people find the instructions for installing sc3-plugins difficult to follow. 

Some quick suggestions for improvements:

1. make it so "release downloads" isn't a separate section, i.e. put download instructions as part of 'binary install'. Then people will be less likely press the green button and  end up trying to install the source with the binary instructions.
2. explicitly telling people not to press the green button is probably a good idea, most people seem to make this mistake at the moment. Or perhaps moving the install instructions off github itself which comes with a lot of confusing information + ui.
3. explaining that the install folder probably won't already exist, and needs creating.
4. explaining what to do if the part of the path is a hidden folder - I think this is the case under macos which makes things very difficult!
5. telling people how to evaluate by putting their cursor on the line and hitting shift-enter or whatever
6. maybe not call the section 'binary install' but just 'install', for those who don't know what 'a binary' means
7. make the link text 'download latest version' or something rather than 'github release page'

Or perhaps sc3-plugins should be rolled into the main release. The package is small by today's standards and I get the impression that most if not all people install them.

brianlheim:
Fixes #176.

- merge "Release downloads" and "binary install"
- rename section "Installation"
- add info about how to evaluate code in SC
- add instructions about how to get to extensions dir and that you might need to create the folder

@yaxu I didn't change the text of the "GitHub release page" link. I'm not sure why that is confusing, and it also doesn't make grammatical sense to change it to "download latest version". If you can explain that I'd appreciate it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,24 +12,34 @@ To learn how to write your own plugins, see [example-plugins](https://github.com
 
 For community discussion and support see the [SuperCollider mailing lists](http://www.birmingham.ac.uk/facilities/BEAST/research/supercollider/mailinglist.aspx) and the [github issue tracker](https://github.com/supercollider/sc3-plugins/issues).
 
-## Release Downloads
+## Installation
 
-Compiled releases are available from the [GitHub release page](https://github.com/supercollider/sc3-plugins/releases). For older versions (2013), see the [Sourceforge project page](http://sourceforge.net/projects/sc3-plugins/files/).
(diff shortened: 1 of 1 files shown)


yaxu:
This is great, thanks @brianlheim. You can ignore the bit about 'download latest version', I agree it probably wouldn't be an improvement.

brianlheim:
Great, thanks again for the suggestions!

PR #183, opened 2018-02-02
VBAP: Fix stairstepping during modulation

### LINKED ISSUE #122: VBAP panning arguments are strictly control rate and not interpolated
The azimuth, elevation, and spread arguments to VBAP are time-quantized with no interpolation, which adds noise to the signal when modulated quickly.

snappizz:
this fixes #122 and replaces #123. i'm kicking myself because i delayed on this an entire year and only just now realized how simple the bug was.

the problem was a failure to increment the channel gains by the slopes in the nova-simd version of the code.

```
~array = VBAPSpeakerArray(2, [-30, 30, 0, -110, 110]);
~buf = ~array.loadToBuffer;
{ VBAP.ar(5, DC.ar(1), ~buf.bufnum, Line.kr(0, 360, 0.05), 1.0, 0.0) }.plot(0.05);
```

before:

vbap_before

after:

vbap_after


muellmusik:
I'll combine with the VBAP 1.0.3.2 stuff etc. when I have time later. A bit bigger job...

Issue #134, opened 2016-12-27
For a release, devs need to manually create source-code tarballs for sc3-plugins


kristho91:
I'm getting following error when trying to build from the release tarball:

```
CMake Error at CMakeLists.txt:6 (message):
nova-simd submodule is missing: please run `git submodule init && git
submodule update' from the toplevel of your git working tree
```

This shouldn't be necessary, right? Or am I using the wrong tarball? (I am using this one: https://github.com/supercollider/sc3-plugins/archive/Version-3.8.0.tar.gz )

vivid-synth:
@kristho91 yes, this seems wrong. In the meantime you can

```
$ git pull https://github.com/supercollider/sc3-plugins
$ cd sc3-plugins
$ git submodule update --init --recursive
```

danstowell:
It's Github's fault. Github creates those tar.gz files automatically and github doesn't include submodules even though it should - see [this issue discussed on the main supercollider codebase](https://github.com/supercollider/supercollider/issues/1807). It is not the fault of the supercollider developers, and the supercollider developers have no way of removing those defective tarballs.

kristho91:
@danstowell I know :) But could you maybe create a tarball that contains those submodules - like supercollider does on their own website?

gusano:
@kristho91 Those tarballs (`Source code (zip)` and `Source code (tar.gz)`) are automagically created by github when creating a release and there's nothing we can do about it, as explained in [the issue Dan mentioned](https://github.com/supercollider/supercollider/issues/1807). All other tarballs are updated manually and contain the submodules.

LFSaw:
we should

1. file an issue for github
2. mention it in the README?

dvzrv:
Implemented with #186

PR #186, opened 2018-02-25
Create assets with a script

### LINKED ISSUE #184: external libraries not included in release!
Hey there,

I'd like to move sc3-plugins to Arch Linux' [community] repository. However, I'm unable to do that, as the external libraries in the release tarball are missing and I can't build the plugins!
Please re-release the tarball.

Without version tags, it is impossible for me to ship sc3-plugins' dependencies either (good packaging style disallows this).

dvzrv:
This is an attempt to fix #184 by creating a script, that can be called from anywhere.
It will use `/tmp` to create assets (in the form of `sc3-plugins-Version-x.x.x.tar.gz`) and then place the finished tarball beneath `package/`.

There is one pitfall: It uses `readlink -f` (see [get_absolute_path()](https://github.com/supercollider/sc3-plugins/compare/master...dvzrv:package?expand=1#diff-07e7b66d35e7dabd5131935313a81d3aR8)), which doesn't seem to be around under macOS by default, but can be installed with the coreutils package using homebrew or macports.

If you know of a valid alternative, let me know. As a workaround the script currently moves the tarball to $HOME on macOS and the package folder on Linux.

IMHO this is a much cleaner solution, than doing this in a potential working copy directly (as it is done in the SuperCollider repository).

### DIFF
--- a/package/create_package.sh
+++ b/package/create_package.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # Creates assets for sc3-plugins in the form of
-# 'sc3-plugins-Version-x.x.x.tar.gz' and moves the file to $HOME (macOS) the
-# repository's package folder (Linux).
+# 'sc3-plugins-Version-x.x.x.tar.gz' and moves the file to the repository's
+# package folder.
 # Requires a writable /tmp folder.
 
 set -euo pipefail
@@ -63,22 +63,33 @@ cleanup_source_dir() {
   rm -rf "${package_name}-Version-${version}"
 }
 
+print_help() {
+  echo "Usage: $0 -v <version tag>"
+  exit 1
+}
+
 upstream="https://github.com/supercollider/sc3-plugins"
 package_name="sc3-plugins"
 source_dir="/tmp"
 version=`date "+%Y-%m-%d"`
 output_dir=$(get_absolute_path $0)
 
-while getopts ":v:s" Option
-do
-  case $Option in
-    v ) version=$OPTARG
+if [ ${#@} -gt 0 ]; then
+  while getopts 'hv:' flag; do
+    case "${flag}" in
+      h) print_help
+          ;;
+      v) version=$OPTARG
+          ;;
+      *)
+        echo "Error! Try '${0} -h'."
+        exit 1
         ;;
-    s ) package_type="source"
-        ;;
-  esac
-done
-shift $(($OPTIND - 1))
+    esac
+  done
+else
+  print_help
+fi
 
 checkout_project
 checkout_external_libraries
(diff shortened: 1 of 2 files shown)


dvzrv:
@snappizz @brianlheim it would be great if you could give some feedback on this, as I can not test it on macOS.

dvzrv:
I just realized, this will also close #134.

telephon:
maybe this might be a workaround instead of `readlink`?

https://serverfault.com/questions/40144/how-can-i-retrieve-the-absolute-filename-in-a-shell-script-on-mac-os-x/406371#406371

dvzrv:
@telephon thanks a lot for the hint! I have corrected it (strange... searched for so long for a viable multi-architecture salvation...). Feel free to test!
If there are more files to exclude, let me know.

dvzrv:
Had the possibility to test this myself right now on macOS.
This seems to work.
However I'm adding a better help and failover, to make it less error prone to use this script!

dvzrv:
Also, I could add an option to sign the package with a given gpg key, if that's something @snappizz would be interested in doing when creating the assets.
In any case, I'd be happy to have this merged and applied any time now.

dvzrv:
Added non-mandatory function to sign tarballs with a given gpg ID/mail.

package/create_package.sh
```diff
@@ -0,0 +1,119 @@
...
+}
+
+checkout_project() {
+  remove_source_dir
+  echo "Cloning project..."
+  cd "$source_dir"
+  git clone $upstream
+  echo "Checking out version: Version-$version"
+  cd "$package_name"
+  git checkout "Version-$version"
```
gusano: @dvzrv This could be simplified to:
`git clone $upstream --branch "Version-$version" --single-branch --recursive`

That way `checkout_external_libraries()` wouldn't be needed anymore

package/create_package.sh
```diff
@@ -0,0 +1,119 @@
...
+# package folder.
+# Requires a writable /tmp folder.
+
+set -euo pipefail
+
+get_absolute_path() {
+  echo "$(cd "$(dirname "$0")" && pwd -P)"
+}
+
+remove_source_dir(){
```
gusano: @dvzrv space missing here

dvzrv: doesn't break, but is changed for consistency! :)

package/create_package.sh
dvzrv: thanks! that simplfies it a lot! :)

snappizz:
changed my mind -- i'm just going to merge this now and address the requested changes in a separate PR. thanks david!

PR #189, opened 2018-03-21
Fixing wrong check for SUPERNOVA (#185)

### LINKED ISSUE #185: building supernova plugins: is supercollider source really required?
In CmakeLists.txt [supernova is searched for in a path, that assumes the SuperCollider sources are around](https://github.com/supercollider/sc3-plugins/blob/master/CMakeLists.txt#L15).
However, it seems that they are not really required to build the supernova plugins (it is only required to have an installed version of supercollider with supernova around): I removed the if statement as a test and it compiles (also the supernova plugins).

Am I just doing something completely stupid, or is the supernova search path assuming wrongly?

dvzrv:
This should fix #185.
Additionally to applying the existing test to NOVA_DISK_IO (instead of to SUPERNOVA), I removed some tab poisening.


dvzrv:
Also applied proper linebreaks in the long messages (some things with variables are still too long, but it's super messy to fix - for me at least).

dvzrv:
Reverted the 2nd commit (also had a merge conflict doing that. Have hopefully reverted it properly. Please have a look!
Do I have to rebase?

gusano:
@dvzrv It seems that this file still has a mix of 2 vs 4 spaces for indention, as well as a couple of TABS still laying around.

I'd suggest either of the following:
- remove all indention changes as well as multiline string (this can be done later in a separate PR) and only keep the supernova fix
- fix all indention so it's consistent

dvzrv:
Sorry for the delay. I now rebased and changed the whole file to 2 spaces based indentation.

brianlheim:
For future reference, we prefer 4-space indentation. (see https://github.com/supercollider/supercollider/wiki/Code-style-guidelines#other-files) Technically that only applies to supercollider/supercollider but I see no reason why we should have separate rules for projects under the supercollider GH org.

Also for future reference @gusano @dvzrv I think it's best if we keep formatting changes separate from substantive changes so it's easier to review. There is another reason too, which is that it's easiest for everyone involved if all formatting is changed simultaneously in the project in a single commit. This saves time committing and reviewing, and also makes git history cleaner since there is just one commit for formatting instead of many commits across many files and dates.

Since there have been way too many bumps in this PR and it's nobody's fault in particular I'm just going to move your work (minus the unrelated formatting) into a new PR and close this one. Thanks for your patience!

brianlheim:
(#229)

PR #193, opened 2018-03-28
Add NHHall reverb UGen

nhthn:
a [new algorithmic reverb](https://github.com/snappizz/nh-ugens) i'm developing. it has low CPU consumption, true stereo signal path, infinite hold support, a handful of EQ and diffusion parameters to mess with, and a decent sound if i may say so myself. it's also MIT licensed.

as the helpfile warning states, this UGen will probably see a few minor tweaks to its sound in the near future. i still encourage people to use it since these changes won't be drastic and are unlikely to upheave a composition (mostly fiddling with delay line lengths to neutralize artifacts).

### DIFF
--- a/source/NHUGens/sc/NHHall.sc
+++ b/source/NHUGens/sc/NHHall.sc
@@ -36,7 +36,7 @@ NHHall : MultiOutUGen {
         );
     }
 
-    init { arg ... theInputs;
+    init { |... theInputs|
         inputs = theInputs;
         ^this.initOutputs(2, rate);
     }


telephon:
would it be hard to make it multichannel?

joslloand:
> best architecture depends on the flavor of multichannel -- ambisonics? 5.1 surround?

Obviously, Ambisonics is best! ;-)

source/CMakeLists.txt
```diff
@@ -243,7 +244,7 @@ endif()
 # OteyPianoUGens
 BUILD_PLUGIN(OteyPianoUGens "OteyPianoUGens/OteyPiano.cpp;OteyPianoUGens/dwgs.cpp;OteyPianoUGens/filter.cpp;OteyPianoUGens/hammer.cpp;OteyPianoUGens/reverb.cpp"
              "" "OteyPianoUGens")
-             
+
```
brianlheim: could you please revert these whitespace changes? IIRC might conflict with #146 if merged in this state.

snappizz: :ok_hand:

patrickdupuis:
@snappizz What's the status of this PR? Are we ready to merge?

telephon:
@snappizz sorry for the silence, I had overlooked your question.

>would you be more interested in a mono-to-multichannel or a multichannel-to-multichannel reverb?

I thought that stereo input is rather particular (but OK!).  Maybe the UGen should be called `NHHall2` to indicate this.

If you can separate subsets of your implementation as separate UGens, then one can just build one's own multichannel version in sclang, maybe?

source/NHUGens/sc/NHHall.sc
```diff
@@ -0,0 +1,45 @@
...
+            hiFreq,
+            hiRatio,
+            earlyDiffusion,
+            lateDiffusion,
+            modRate,
+            modDepth
+        );
+    }
+
+    init { arg ... theInputs;
```
telephon: better use `|...theInputs|` when you use them above, too.

snappizz:
the only serious issue with this UGen is a denormal number bug with ARM processors (rpi, bela, etc.). i've been working with patrick to try to track it down.

other than that, i'm dissatisfied with a weird stereo "wobble" in the initial onset of the reverb tail, so some tweaks will be necessary. both of these issues are documented in the helpfile, so this should be good to go.

> I thought that stereo input is rather particular (but OK!). Maybe the UGen should be called NHHall2 to indicate this.

thanks for the suggestion, but i think the name is fine. the vast majority of commercial reverbs are stereo as well.

source/NHUGens/sc/NHHall.sc
snappizz: :ok_hand:

Issue #191, opened 2018-03-21
"Policy CMP0084 is not set" (StkInst)


dvzrv:
I get this, when building sc3-plugins:

```
CMake Warning (dev) at source/StkInst/CMakeLists.txt:3 (project):
  Policy CMP0048 is not set: project() command manages VERSION variables.
  Run "cmake --help-policy CMP0048" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  The following variable(s) would be set to empty:

    PROJECT_VERSION
    PROJECT_VERSION_MAJOR
    PROJECT_VERSION_MINOR
    PROJECT_VERSION_PATCH
This warning is for project developers.  Use -Wno-dev to suppress it.

```

brianlheim:
cc @hzulla @danstowell @sonoro1234 who are the people who seem to have edited that file most.

sonoro1234:
It seems that from cmake 3.02 version has to be set in project according to
```
https://cmake.org/cmake/help/v3.0/command/project.html#command:project
```
perhaps setting VERSION 1 would stop warnings

sonoro1234:
@brianlheim It is solved in PR 196

brianlheim:
@sonoro1234 - I've already seen your PR and will review it when I have time.

PR #196, opened 2018-05-23
stop CMake warning in StkInst issue#191

sonoro1234:


sonoro1234:
solves issue #191

simonvanderveldt:
It would probably be cleaner to actually make use of the variables provided by `project()` like `PROJECT_NAME` and `PROJECT_VERSION` instead of entirely getting rid of it.
See https://cmake.org/cmake/help/v3.0/command/project.html#command:project

One strange thing, the main `CMakeLists.txt` also [uses `project()`](https://github.com/supercollider/sc3-plugins/blob/6a039c4f0d450129c36db1ffc744117beb115218/CMakeLists.txt#L3), doesn't have a `cmake_policy` command but doesn't throw this warning. Not sure why that is.

sonoro1234:
Setting version results in
```
CMake Error at source/StkInst/CMakeLists.txt:3 (project):
  VERSION not allowed unless CMP0048 is set to NEW
```
Yes, it is strange that the error doesnt happen in other places!!

brianlheim:
> It would probably be cleaner and more future proof to make use of the variables provided by project() like PROJECT_NAME and PROJECT_VERSION instead of entirely getting rid of it.
See https://cmake.org/cmake/help/v3.0/command/project.html#command:project

I agree. Just pass it `VERSION ${PROJECT_VERSION}`.

> One strange thing, the main CMakeLists.txt also uses project(), doesn't have a cmake_policy command but doesn't throw this warning. Not sure why that is.

See `cmake --help-policy CMP0048`.

> Setting version results in
> CMake Error at source/StkInst/CMakeLists.txt:3 (project):
>   VERSION not allowed unless CMP0048 is set to NEW

Then set CMP0048 to new, it was introduced in 3.0 so we are fine (min version should be updated to 3.5 at this point, that is an oversight)

brianlheim:
set it explicitly and in the top directory.

sonoro1234:
done

dyfer:
closing and re-opening to re-trigger travis

brianlheim:
waiting for appveyor before merge

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
-cmake_minimum_required (VERSION 2.8)
-set(SUPERNOVA_CMAKE_MINVERSION 3.1)
-project (sc3-plugins)
+cmake_minimum_required (VERSION 3.5)
+cmake_policy(SET CMP0048 NEW)
+project (sc3-plugins VERSION ${PROJECT_VERSION})
```
dyfer, line 3: It seems that in main CMakeLists we now use `${PROJECT_VERSION}` before defining if further down, resulting in a message from `cmake`:
```sh
CMake Warning at CMakeLists.txt:3 (project):
  VERSION keyword not followed by a value or was followed by a value that
  expanded to nothing.
```
Not sure what the proper solution is, but moving the section
```cmake
include("${SC_PATH}/SCVersion.txt")
set(PROJECT_VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}${PROJECT_VERSION_PATCH}")
message(STATUS "Building plugins for SuperCollider version: ${PROJECT_VERSION}")
```
before this line worked for me. @sonoro1234 @brianlheim what do you think?

brianlheim: that is probably good

brianlheim:
the merge commit ("merge master into cmakewarning") needs to be redone as a rebase and not merge before this can move forward

CMakeLists.txt
brianlheim: if it silences the warning and doesn't seem to cause any obvious issues i think that's fine to ship. thanks for looking into it @dyfer :)

dyfer:
@sonoro1234 would you make the suggested change and then rebase? thanks!

sonoro1234:
> would you make the suggested change and then rebase? thanks!

There are two problems:
1- I am very bad at git and dont know what rebase is
2- I am not sure to understand the changes requested

brianlheim:
@sonoro1234 https://github.com/supercollider/supercollider/wiki/Creating-pull-requests#notes-on-rebasing-and-merge-conflicts

and remove the merge commit from the list of commits when the editor window appears

brianlheim:
then after you do that, the other "suggested change" is @dyfer 's review comment

sonoro1234:
Did a new PR  with suggested changes

sonoro1234:
closing to aviod confussion

PR #198, opened 2018-06-19
Improve HOAUGens README

patrickdupuis:
This PR slightly improves the HOAUGens README with regards to building the UGens. Some members have expressed concerns with the README in its current state. Please make suggestions on what is missing from this text and I will add the necessary info.


source/HOAUGens/sc/README.md
```diff
@@ -16,20 +16,20 @@ The SuperCollider library was written by [Florian Grond](http://www.grond.at). I
 
 ## Installation
 
-The HOAUGens can be compiled and installed as part of the sc3-plugins collection. Compilation of the HOAUGens can be turned `ON` by setting the corresponding cmake variable:
+The HOAUGens can be installed as part of the sc3-plugins collection. In order for the HOAUGens to be built when compiling sc3-plugins, the cmake variable `HOA_UGENS` must be set to `ON`.
```
jamshark70: Could be a bit tighter:

> The HOAUGens can be installed as part of the sc3-plugins collection. Because of compiler restrictions on specific platforms (below), HOAUGens are currently disabled by default. To enable them, set the cmake variable `HOA_UGENS` to `ON`, as follows:

Then the subsequent paragraph can be trimmed or removed.

brianlheim: This is also my only comment. the rest looks good, thanks.

brianlheim:
I added some info to the main README about HOAUGens, thanks for the suggestion @jamshark70 . If all looks good I'm ok merging this now.

patrickdupuis:
I'm guessing this is OK to merge regardless of Travis?

brianlheim:
Yep, travis got confused about which branch to check

PR #213, opened 2018-09-19
Add QuantityUGens, MovingSum and MovingAverage

mtmccrea:
Add QuantityUGens, intended for analyzing signal/soundfield quantities. Currently just adds
MovingSum, which is a variation of RunningSum that includes a variable window size. Extended to time averaging by MovingAverage class and convenience methods *power and *rms. Help docs and unit test included.

### DIFF
--- a/source/QuantityUGens/QuantityUGens.cpp
+++ b/source/QuantityUGens/QuantityUGens.cpp
@@ -1,22 +1,22 @@
 /*
-	SuperCollider real time audio synthesis system
-    Copyright (c) 2002 James McCartney. All rights reserved.
-	http://www.audiosynth.com
-
-    This program is free software; you can redistribute it and/or modify
-    it under the terms of the GNU General Public License as published by
-    the Free Software Foundation; either version 2 of the License, or
-    (at your option) any later version.
-
-    This program is distributed in the hope that it will be useful,
-    but WITHOUT ANY WARRANTY; without even the implied warranty of
-    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-    GNU General Public License for more details.
-
-    You should have received a copy of the GNU General Public License
-    along with this program; if not, write to the Free Software
-    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
-*/
+ SuperCollider real time audio synthesis system
+ Copyright (c) 2002 James McCartney. All rights reserved.
+ http://www.audiosynth.com
+ 
+ This program is free software; you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation; either version 2 of the License, or
+ (at your option) any later version.
+ 
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+ 
+ You should have received a copy of the GNU General Public License
+ along with this program; if not, write to the Free Software
+ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
+ */
 
 
 /*
@@ -28,30 +28,52 @@
  Authors:
  Michael McCrea, 2018, mtm5@uw.edu
  
-*/
+ */
 
 #include "SC_PlugIn.h"
 
 static InterfaceTable *ft;
 
-
-// MovingSum uses the underlying strategy of RunningSum
-// but adds a variable window size and a flag to average the output.
-struct MovingSum : public Unit {
+// MovingSum/Average uses the underlying strategy of RunningSum
+// but adds a variable window size.
+struct MovingX : public Unit {
     int nsamps, maxSamps, head, tail, resetCounter;
     float msum, msum2;
     bool reset;
     float* msquares;
 };
 
+struct MovingSum : public MovingX
+{};
+
+struct MovingAverage : public MovingX
+{};
+
 extern "C"
 {
     void MovingSum_Ctor(MovingSum *unit);
     void MovingSum_Dtor(MovingSum *unit);
     void MovingSum_next(MovingSum *unit, int inNumSamples);
+    
+    void MovingAverage_Ctor(MovingAverage *unit);
+    void MovingAverage_Dtor(MovingAverage *unit);
+    void MovingAverage_next(MovingAverage *unit, int inNumSamples);
+    
+    void MovingXCtorHelper(MovingX *unit, UnitCalcFunc calcFunc);
+    void MovingX_next(MovingX *unit, int inNumSamples, bool avg);
 }
 
 void MovingSum_Ctor( MovingSum* unit )
+{
+    MovingXCtorHelper(unit, (UnitCalcFunc)&MovingSum_next);
+}
+
+void MovingAverage_Ctor( MovingAverage* unit )
+{
+    MovingXCtorHelper(unit, (UnitCalcFunc)&MovingAverage_next);
+}
+
+inline void MovingXCtorHelper( MovingX* unit, UnitCalcFunc calcFunc )
 {
     if ((int) ZIN0(2) == 0) {
         printf("MovingSum Error: maxSamps initialized to 0.\n");
@@ -59,11 +81,11 @@ void MovingSum_Ctor( MovingSum* unit )
         unit->mDone = true;
         return;
     }
-
-    SETCALC(MovingSum_next);
-
+    
+    unit->mCalcFunc = calcFunc;
+    
     unit->maxSamps  = (int) ZIN0(2);
-    unit->nsamps    = sc_max(1, sc_min((int) ZIN0(1), unit->maxSamps)); // clip(1, maxSamps)
+    unit->nsamps    = sc_clip((int) ZIN0(1), 1, unit->maxSamps);
     unit->msum      = 0.0f;
     unit->msum2     = 0.0f;
     unit->resetCounter = 0;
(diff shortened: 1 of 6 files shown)


source/QuantityUGens/QuantityUGens.cpp
```diff
@@ -0,0 +1,212 @@
...
+    RTFree(unit->mWorld, unit->msquares);
+}
+
+void MovingSum_next( MovingSum *unit, int inNumSamples )
+{
+    float *in   = ZIN(0);
+    float *out  = ZOUT(0);
+    float *data = unit->msquares;
+	
+	int newWinSize  = (int) ZIN0(1);  // number of samples to sum
```
joshpar: please update for consistency with tabs / spaces.

brianlheim: (use 4 spaces)

source/QuantityUGens/QuantityUGens.cpp
```diff
@@ -0,0 +1,212 @@
...
+        printf("MovingSum Error: maxSamps initialized to 0.\n");
+        SETCALC(*ClearUnitOutputs);
+        unit->mDone = true;
+        return;
+    }
+
+    SETCALC(MovingSum_next);
+
+    unit->maxSamps  = (int) ZIN0(2);
+    unit->nsamps    = sc_max(1, sc_min((int) ZIN0(1), unit->maxSamps)); // clip(1, maxSamps)
```
brianlheim: there is `sc_clip` for this

source/QuantityUGens/sc/HelpSource/Classes/MovingSum.schelp
```diff
@@ -0,0 +1,108 @@
...
+Size of the summing window, in samples. Modulatable between 1 and
+strong::maxsamp::. Default: 40.
+argument::maxsamp
+Maximum size of the summing window. Pre-allocates memory up to this size and
+your strong::numsamp:: will be clipped at this value. strong::maxsamp:: cannot
+be modulated.
+argument::avg
+A flag specifying whether to output the average of the running sum over the
+strong::numsamp:: frames. code::0:: outputs the sum, code::> 0:: outputs the
+average. This is used internally by link::Classes/MovingAverage::.
```
brianlheim: Why leave this public/configurable? It seems like you could very easily refactor the plugin so that you have both a real `MovingSum` and `MovingAvg`, and not have this semantic leak where "MovingSum" is actually "MovingSumOrAverage"

mtmccrea: Ah, keeping me honest, thanks ;)
OK I've refactored it, but as C++ isn't really my sandbox, I could use a second pair of eyes on the way I did it to make sure I'm not missing a cleaner approach.
I basically confined the flag to average the output to the main calculation function, which is shared between both UGens. 
I've also created a shared constructor between them, MovingXCtorHelper, and the _next calculation function is passed in from the constructor proper (_Ctor). This is where it would be helpful to double check the way the _next  functions are passed in.

source/QuantityUGens/QuantityUGens.cpp
```diff
@@ -0,0 +1,248 @@
...
+}
+
+void MovingSum_Ctor( MovingSum* unit )
+{
+    MovingXCtorHelper(unit, (UnitCalcFunc)&MovingSum_next);
+}
+
+void MovingAverage_Ctor( MovingAverage* unit )
+{
+    MovingXCtorHelper(unit, (UnitCalcFunc)&MovingAverage_next);
```
mtmccrea: @brianlheim how does this look for passing around the calculation function?

mtmccrea: I tried it a number of other ways, and couldn't get it to work, so I followed the pattern I saw in the Delay UGens:
https://github.com/supercollider/supercollider/blob/f65a1ef7bfb100f4ccb7dcab3a6dff6826be4344/server/plugins/DelayUGens.cpp#L3039

source/QuantityUGens/QuantityUGens.cpp
mtmccrea: Hi @joshpar, I've addressed the whitespace issue, and it looks like this comment is still blocking. Would you mind reviewing?

joshpar:
@brianlheim - any other changes you want on your end? IF not, I can merge.

source/QuantityUGens/QuantityUGens.cpp
```diff
@@ -0,0 +1,248 @@
...
+}
+
+inline void MovingXCtorHelper( MovingX* unit, UnitCalcFunc calcFunc )
+{
+    if ((int) ZIN0(2) == 0) {
+        printf("MovingSum Error: maxSamps initialized to 0.\n");
```
telephon: to be safe, shouldn't this be `ZIN0(2) <= 0` ?

mtmccrea: Yes, thanks @telephon, I've updated it.

source/QuantityUGens/QuantityUGens.cpp
```diff
@@ -75,8 +75,8 @@ void MovingAverage_Ctor( MovingAverage* unit )
 
 inline void MovingXCtorHelper( MovingX* unit, UnitCalcFunc calcFunc )
 {
-    if ((int) ZIN0(2) == 0) {
-        printf("MovingSum Error: maxSamps initialized to 0.\n");
+    if ((int) ZIN0(2) <= 0) {
```
brianlheim: that was a good catch @telephon . also FYI this is undefined behavior when the value is +/-inf, NaN, or has an integral part larger than INT_MAX. whether or not you want to care about those cases or just let it blow up is up to you.

source/QuantityUGens/QuantityUGens.cpp
brianlheim: it looks fine, i'm not able to think of an idiomatic way to do this in C. you have a pretty strong guarantee that these functions are only going to be passed pointers to the right types anyway.

brianlheim:
Sorry for the delay @mtmccrea . I am fine merging this as is but you might want to consider the comment I left above.

mtmccrea:
@brianlheim thanks for bringing this up. I've amended the commit to include tests for inf and nan, as well as `< INT_MAX`. It's rather brute force, but after looking around I didn't see these checks commonly handled together, I'm wondering if there's a catch-all similar to `zapgremlins` (could/should this be used here?)

I also noticed that while testing for these cases, the catch worked, but the synth containing the UGen still played, outputting 0 (which I think is correct behavior), but `cmd + .` killed the server (`RESULT = 0 \  server 'localhost' disconnected shared memory interface`. I noticed that this isn't the case when running the synth/UGen with the argument within bounds. This seems like a red flag, though I'm not sure why this is happening. Ideas?

brianlheim:
> It's rather brute force, but after looking around I didn't see these checks commonly handled together, I'm wondering if there's a catch-all similar to zapgremlins (could/should this be used here?)

a generic `template<typename Integral, typename Floating> [constexpr] bool canBeRepresentedAs(Floating f);` could be very useful. Since literally all of our code is compiled in C++11 at minimum I'd rather utilize std::numeric_limits than handcode all the cases in C.

> This seems like a red flag, though I'm not sure why this is happening.
> Ideas?

Start with the core dump. Are you trying to `free` a default-initialized (i.e. garbage) pointer in the destructor?

source/QuantityUGens/QuantityUGens.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+
+void MovingSum_Dtor(MovingSum *unit)
+{
+    RTFree(unit->mWorld, unit->msquares);
+}
+void MovingAverage_Dtor(MovingAverage *unit)
+{
+    RTFree(unit->mWorld, unit->msquares);
```
brianlheim: If `unit->msquares` is default initialized (and I think ugens are) this is UB which might cause your crash. You could always assign it to nullptr at the top of the ctor and then check here.

mtmccrea:
Thanks @brianlheim, both of these suggestions were helpful!

Catching `nullptr` did the trick to prevent the crash on `cmd + .`

Conveniently, `std::isfinite()` checks for both `inf` and `NaN`, so that reduced the number of checks. I missed this recommendation in the [Server Plugin API](http://doc.sccode.org/Reference/ServerPluginAPI.html#Unary%20operators).

Thanks again, these will be useful for writing UGens in the future...

mtmccrea:
I missed your comment about the type promotion in the input test, following up shortly..

source/QuantityUGens/QuantityUGens.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+void MovingAverage_Ctor( MovingAverage* unit )
+{
+    MovingXCtorHelper(unit, (UnitCalcFunc)&MovingAverage_next);
+}
+
+inline void MovingXCtorHelper( MovingX* unit, UnitCalcFunc calcFunc )
+{
+	if (((int) ZIN0(2) <= 0) || ((int) ZIN0(2) > INT_MAX) || sc_isnan(ZIN0(2)) || !sc_isfinite(ZIN0(2))) {
```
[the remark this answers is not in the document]
mtmccrea: Thanks for the tip on type promotion, I've updated the checks.
The indentation...ugh, sorry. Because this file doesn't belong to a project I have to remember to set the indentation every time I open the file in Xcode. Because I don't use it often, I haven't yet developed the habit, but I'm showing invisibles now so that should help. Noted!

source/QuantityUGens/QuantityUGens.cpp
```diff
@@ -115,11 +117,15 @@ inline void MovingXCtorHelper( MovingX* unit, UnitCalcFunc calcFunc )
...
+    if (unit->msquares) {
+        RTFree(unit->mWorld, unit->msquares);
+    }
 }
 void MovingAverage_Dtor(MovingAverage *unit)
 {
-    RTFree(unit->mWorld, unit->msquares);
+    if (unit->msquares) {
+        RTFree(unit->mWorld, unit->msquares);
+    }
```
brianlheim: I wasn't sure about this when i made my comment earlier but it turns out it's OK to pass RTFree nullptr, FYI.

brianlheim: https://github.com/supercollider/supercollider/blob/f65a1ef7bfb100f4ccb7dcab3a6dff6826be4344/common/SC_AllocPool.cpp#L164

mtmccrea:
Thanks @brianlheim for keeping a close eye on it, I appreciate the learning opportunity :)

brianlheim:
@mtmccrea Of course, glad to help. :)

PR #220, opened 2018-10-10
[travis] add build for linux

### LINKED ISSUE #218: Travis builds for linux
Issues like #216 could be avoided if we had travis building for linux and not for osx only.
I could do that quickly this week.

gusano:
This enables linux builds on travis and fixes #218:

- avoid installing unnecessary packages
- use travis default packages (`gcc`, `cmake`, `build-essentials`, ..)
- osx defaults to `Xcode-9.4.1`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,33 +1,31 @@
-language: objective-c
+language: cpp
 
-os:
- - osx
+matrix:
+  fast_finish: true
+  include:
+    - os: linux
+      sudo: required
+      dist: trusty
+    - os: osx
 
 cache:
+ - ccache
  - apt
  - bundler
 
 before_install:
  - ifmac () { if [[ $TRAVIS_OS_NAME == osx ]]; then eval $@; fi; }
  - iflin () { if [[ $TRAVIS_OS_NAME == linux ]]; then eval $@; fi; }
- - ifmac brew install cmake || true
- - iflin sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
- - iflin sudo add-apt-repository -y ppa:andykimpe/cmake  # backport of cmake 2.8.12
- - iflin sudo add-apt-repository -y ppa:ondrej/php5 # libicu-dev 52
- - iflin sudo apt-get update
- - iflin sudo apt-get install libicu-dev=52.1-1+debphp.org~precise+1 gcc-4.7 g++-4.7 aptitude build-essential libfftw3-dev libxt-dev pkg-config cmake=2.8.12.2-3 libstdc++5
- - iflin sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.6 40 --slave /usr/bin/g++ g++ /usr/bin/g++-4.6
- - iflin sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.7 60 --slave /usr/bin/g++ g++ /usr/bin/g++-4.7
- - iflin sudo update-alternatives --auto gcc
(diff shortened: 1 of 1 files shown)


gusano:
bump?

dvzrv:
It could be helpful to add an additional build. that tries `SYSTEM_STK`, otherwise: LGTM

.travis.yml
```diff
@@ -1,33 +1,31 @@
...
- - iflin sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.7 60 --slave /usr/bin/g++ g++ /usr/bin/g++-4.7
- - iflin sudo update-alternatives --auto gcc
+ - iflin sudo apt-get install libicu-dev libfftw3-dev libxt-dev pkg-config libstdc++5
  - git clone https://github.com/supercollider/supercollider.git
 
 before_script:
- - mkdir BUILD
- - cd BUILD
+ - mkdir BUILD && cd BUILD
  - cmake -DCMAKE_INSTALL_PREFIX:PATH=$PWD/SC3plugins -DCMAKE_BUILD_TYPE=Release -DSC_PATH=../supercollider ..
```
dvzrv: The only thing that could potentially be interesting here is to also try a variant where we use `SYSTEM_STK`. In trusty stk is around: https://packages.ubuntu.com/trusty/stk (I tried and build with 4.6.0 on Arch Linux).

.travis.yml
```diff
@@ -1,33 +1,31 @@
...
- - ifmac brew install cmake || true
- - iflin sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
- - iflin sudo add-apt-repository -y ppa:andykimpe/cmake  # backport of cmake 2.8.12
- - iflin sudo add-apt-repository -y ppa:ondrej/php5 # libicu-dev 52
- - iflin sudo apt-get update
- - iflin sudo apt-get install libicu-dev=52.1-1+debphp.org~precise+1 gcc-4.7 g++-4.7 aptitude build-essential libfftw3-dev libxt-dev pkg-config cmake=2.8.12.2-3 libstdc++5
- - iflin sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.6 40 --slave /usr/bin/g++ g++ /usr/bin/g++-4.6
- - iflin sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.7 60 --slave /usr/bin/g++ g++ /usr/bin/g++-4.7
- - iflin sudo update-alternatives --auto gcc
+ - iflin sudo apt-get install libicu-dev libfftw3-dev libxt-dev pkg-config libstdc++5
```
dvzrv: Adding stk for the additional build with `SYSTEM_STK`

.travis.yml
```diff
@@ -1,33 +1,31 @@
...
+language: cpp
 
-os:
- - osx
+matrix:
+  fast_finish: true
+  include:
+    - os: linux
+      sudo: required
+      dist: trusty
```
dvzrv: This could be extended to something more probing (using different GCC versions, as the default is quite old). For an example: https://github.com/SoundScapeRenderer/ssr/blob/master/.travis.yml

gusano: @dvzrv The scope of this PR is to provide at least one basic linux build setup so we don't hit the issue we hit last month or so about plugins not building on linux in general (right now there is no check at all).
Testing `STK` and other `gcc` versions sounds great but maybe in another PR?

ps: LGTM is not enough, PR needs to be approved : )

dvzrv: Oh, my bad then! :) Go ahead! :)

PR #223, opened 2018-10-11
set first sample of output in Ctor

sonoro1234:
possibly solves #214

Also need that in DWGBowed , DWGPlucked and OteyPianoUGens. Shall I wait for this commited or include changes in this PR?


jamshark70:
I'd say, include those fixes here. If the issue and solution are known, there's no need for a separate PR.

Thanks!

source/StkInst/StkInst.cpp
```diff
@@ -411,6 +412,7 @@ void StkGlobals_Ctor(StkGlobals* unit) {
     Stk::printErrors(unit->printErrors);
     RTFree(unit->mWorld,unit->rawwavePath);
     SETCALC(StkGlobals_next);
+    (unit->mCalcFunc)(unit, 1);
```
snappizz: why call calcfunc in the Dtor?

snappizz:
(deleted a comment which was based on a misunderstanding)

sonoro1234:
done

sonoro1234:
Is somebody going to review?

sonoro1234:
3.10.0-rc1?

source/OteyPianoUGens/OteyPiano.cpp
```diff
@@ -328,6 +329,7 @@ void OteyPianoStrings_Ctor(OteyPianoStrings* unit) {
 	new(unit) OteyPianoStrings(unit);
 	unit->piano.init(freq,SAMPLERATE,velocity*10,minr,maxr,amprl,amprr,rcore,minl,maxl,ampll,amplr,rho,young,zb,zh,mh,k,alpha,p,ph,loss,detune,hammer_type);
 	SETCALC(OteyPianoStrings_next);
+	ZOUT0(0) = 0.0;
```
snappizz: `ZOUT0` and all other `Z*` macros are legacy. please use `ClearUnitOutputs(unit, 1)` here and elsewhere.

sonoro1234:
Done

PR #222, opened 2018-10-10
[jekyll] update gems

gusano:
Fix #221

PR #228, opened 2018-11-25
update site for 3.10

brianlheim:
fix download link, fix typos
update Gemfile.lock


gusano:
@brianlheim I guess this makes #222 obsolete

PR #229, opened 2018-11-30
CMake: Fix wrong check for SUPERNOVA

### LINKED ISSUE #185: building supernova plugins: is supercollider source really required?

brianlheim:
Fixes #185

so default builds won't fail if SC sources are not around during build time

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,19 +22,6 @@ if (NOT SYSTEM_STK)
   endif()
 endif()
 
-set(NOVA_TT_MISSING_ERROR "The nova-tt source code is missing in \
-${SC_PATH}/external_libraries/nova-tt.\n Make sure to point to a valid version \
-of SuperCollider's source code (with the help of the SC_PATH variable).\n It's \
-easiest to use a release tarball of SuperCollider.\n If you use a git clone, \
-make sure you have all submodules by running `git submodule update --init \
---recursive` in the root of SuperCollider's source code repository.")
-
-if (NOVA_DISK_IO)
-  if (NOT EXISTS ${SC_PATH}/external_libraries/nova-tt/CMakeLists.txt)
-    message(FATAL_ERROR ${NOVA_TT_MISSING_ERROR})
-  endif()
-endif()
-
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake_modules
                       ${CMAKE_MODULE_PATH})
 
@@ -51,6 +38,19 @@ if (NOT SC_FOUND)
 	message(SEND_ERROR "cannot find SuperCollider3 headers. Set the variable SC_PATH.")
 else()
 	message(STATUS "Using SC source located at ${SC_PATH}")
+endif()
+
+set(NOVA_TT_MISSING_ERROR "The nova-tt source code is missing in \
+${SC_PATH}/external_libraries/nova-tt.\n Make sure to point to a valid version \
+of SuperCollider's source code (with the help of the SC_PATH variable).\n It's \
+easiest to use a release tarball of SuperCollider.\n If you use a git clone, \
+make sure you have all submodules by running `git submodule update --init \
+--recursive` in the root of SuperCollider's source code repository.")
+
+if (NOVA_DISK_IO)
+  if (NOT EXISTS ${SC_PATH}/external_libraries/nova-tt/CMakeLists.txt)
+    message(FATAL_ERROR ${NOVA_TT_MISSING_ERROR})
+  endif()
 endif()
 
 include("${SC_PATH}/SCVersion.txt")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,19 +2,36 @@ cmake_minimum_required (VERSION 2.8)
 set(SUPERNOVA_CMAKE_MINVERSION 3.1)
 project (sc3-plugins)
 
+set(NOVA_SIMD_MISSING_ERROR "The nova-simd source code is missing in \
+${CMAKE_CURRENT_SOURCE_DIR}/external_libraries/nova-simd.\n This probably \
+means you forgot to clone submodules.\n To fix this, run `git submodule \
+update --init` from the root of the sc3-plugins repository")
+
 if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/external_libraries/nova-simd/simd_memory.hpp)
-  message(FATAL_ERROR "The nova-simd submodule is missing. This probably means you forgot to clone submodules. To fix this, run `git submodule update --init` from the root folder of sc3-plugins repository")
+  message(FATAL_ERROR "${NOVA_SIMD_MISSING_ERROR}")
 endif()
+
+set(STK_MISSING_ERROR "The stk source code is missing in \
+${CMAKE_CURRENT_SOURCE_DIR}/external_libraries/stk.\n This probably means you \
+forgot to clone submodules.\n To fix this, run `git submodule update --init` \
+from the root of the sc3-plugins repository")
 
 if (NOT SYSTEM_STK)
   if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/external_libraries/stk/include/Stk.h)
-    message(FATAL_ERROR "The stk submodule is missing. This probably means you forgot to clone submodules. To fix this, run `git submodule update --init` from the root folder of the sc3-plugins repository")
+    message(FATAL_ERROR ${STK_MISSING_ERROR})
   endif()
 endif()
 
-if (SUPERNOVA)
+set(NOVA_TT_MISSING_ERROR "The nova-tt source code is missing in \
+${SC_PATH}/external_libraries/nova-tt.\n Make sure to point to a valid version \
+of SuperCollider's source code (with the help of the SC_PATH variable).\n It's \
+easiest to use a release tarball of SuperCollider.\n If you use a git clone, \
+make sure you have all submodules by running `git submodule update --init \
+--recursive` in the root of SuperCollider's source code repository.")
+
+if (NOVA_DISK_IO)
   if (NOT EXISTS ${SC_PATH}/external_libraries/nova-tt/CMakeLists.txt)
-    message(FATAL_ERROR "The nova-tt submodule in the SuperCollider repository is missing (required for SuperNova plugins). This probably means you forgot to clone submodules. To fix this, run `git submodule update --init` from the root folder of the SuperCollider repository")
+    message(FATAL_ERROR ${NOVA_TT_MISSING_ERROR})
   endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -2,19 +2,36 @@ cmake_minimum_required (VERSION 2.8)
...
 project (sc3-plugins)
 
+set(NOVA_SIMD_MISSING_ERROR "The nova-simd source code is missing in \
+${CMAKE_CURRENT_SOURCE_DIR}/external_libraries/nova-simd.\n This probably \
+means you forgot to clone submodules.\n To fix this, run `git submodule \
+update --init` from the root of the sc3-plugins repository")
+
 if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/external_libraries/nova-simd/simd_memory.hpp)
-  message(FATAL_ERROR "The nova-simd submodule is missing. This probably means you forgot to clone submodules. To fix this, run `git submodule update --init` from the root folder of sc3-plugins repository")
+  message(FATAL_ERROR "${NOVA_SIMD_MISSING_ERROR}")
```
patrickdupuis: There's an inconsistent use of quotes around variable names. According to this https://stackoverflow.com/questions/35847655/when-to-quote-variables, we should be using quotes around all of these variables since they contain text.

CMakeLists.txt
```diff
@@ -2,19 +2,36 @@ cmake_minimum_required (VERSION 2.8)
...
   endif()
 endif()
 
-if (SUPERNOVA)
+set(NOVA_TT_MISSING_ERROR "The nova-tt source code is missing in \
+${SC_PATH}/external_libraries/nova-tt.\n Make sure to point to a valid version \
+of SuperCollider's source code (with the help of the SC_PATH variable).\n It's \
+easiest to use a release tarball of SuperCollider.\n If you use a git clone, \
+make sure you have all submodules by running `git submodule update --init \
+--recursive` in the root of SuperCollider's source code repository.")
```
patrickdupuis: This gives:
```
  The nova-tt source code is missing in
  SC_PATH-NOTFOUND/external_libraries/nova-tt.
```
You need to use `${CMAKE_CURRENT_SOURCE_DIR}` here as well :)

CMakeLists.txt
brianlheim: good catch!

CMakeLists.txt
brianlheim: this check should probably go after line 54 (when SC_PATH is guaranteed to be populated)

dvzrv: @brianlheim Ah yeah, that is just in the wrong place. Also, thanks for moving the PR! It got very messy and I try to finish up on it today so this can be merged asap.

@patrickdupuis in this case we're really referencing the SuperCollider source code, so `SC_PATH` is correct.

dvzrv: Probing a little further, I realize, that `SC_PATH` is empty if not specifically setup (`-DSC_PATH=/path/to/supercollider`) - @patrickdupuis I assume you did something similar to just `cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release`, I usually added `-DSC_PATH=/usr/include/SuperCollider` by default.

However, the implicit default, as mentioned in [README.md](https://github.com/supercollider/sc3-plugins/blob/master/README.md#compile-from-source) is not set at all anywhere and leads to an error, when trying to include the `SCVersion.txt` (when I move the `NOVA_TT_MISSING_ERROR` handling below line 54).
This is because in `cmake_modules/FindSuperCollider3.cmake` we probe for [include/plugin_interface/SC_PlugIn.h](https://github.com/supercollider/sc3-plugins/blob/master/cmake_modules/FindSuperCollider3.cmake#L16). Without the **include/** (on Linux) installed SuperCollider files below `/usr/include/SuperCollider` are then detected and also the `SCVersion.txt` gets included as it should be.

Having `SC_PATH` as the single truth for both SuperCollider headers **and** source code is a little confusing and should be revised.

brianlheim:
@dvzrv Still need to address this comment - https://github.com/supercollider/sc3-plugins/pull/229#discussion_r237735782

CMakeLists.txt
dvzrv: Argh, sorry, I jumped over this piece of review!
Yes, those should be quoted (and are now :) ), although in `message()` they really [don't seem to pose any problems](https://cmake.org/cmake/help/v3.13/command/message.html).
However, in a `if(NOT EXIST )` statement they might (if the non-quoted string contains white spaces).

brianlheim:
@patrickdupuis @snappizz - mind reviewing this? I can't as it's my PR

patrickdupuis:
On `master`, if I don't set `SC_PATH`, I get:

```
CMake Error at CMakeLists.txt:34 (message):
  cannot find SuperCollider3 headers.  Set the variable SC_PATH.


CMake Error at CMakeLists.txt:39 (include):
  include could not find load file:

    SC_PATH-NOTFOUND/SCVersion.txt
```

On this branch, I get no such error, and `SC_PATH` is set to `/usr/local/include/SuperCollider`. This means that the `NOVA_DISK_IO` error message looks like this:
```
CMake Error at CMakeLists.txt:52 (message):
  The nova-tt source code is missing in
  /usr/local/include/SuperCollider/external_libraries/nova-tt.

   Make sure to point to a valid version of SuperCollider's source code (with the help of the SC_PATH variable).
   It's easiest to use a release tarball of SuperCollider.
   If you use a git clone, make sure you have all submodules by running `git submodule update --init --recursive` in the root of SuperCollider's source code repository.

```

dvzrv:
@patrickdupuis do you have a version of SuperCollider installed in that location? I hope detection is correct in that case then.
As pointed out before: `SC_PATH` both meaning SuperCollider source and compiled version location is a pretty unfortunate state.
Ideally though, if one wants to compile `NOVA_DISK_IO` `SC_PATH` definitely has to be set to the source code location of SuperCollider - preferably of the same version as the version in use on the system. This is the only Ugen, that really requires the source code of SuperCollider though (because of nova-tt). Maybe it's worth including nova-tt in sc3-plugins' external_libraries?

patrickdupuis:
Think I have a better understanding of this PR now. 

The point I was trying to make was that mentioning `/usr/local/include/SuperCollider/external_libraries/nova-tt` in the error message doesn't make sense. Nova-tt will never be found at that path; it can only be found in the downloaded or git cloned SC source. Maybe the error message doesn't need to print `SC_PATH`? We can simply instruct users on what they need to do in order to build `NOVA_DISK_IO`.

@dvzrv's suggestion of including nova-tt as part of the sc3-plugins submodules sounds to me like something we should consider.

dvzrv:
@patrickdupuis well, this might be taken care of in a separate pull request then.
This pull request at least fixes the wrong check for `SUPERNOVA`, when it should've been `NOVA_DISK_IO`.

dvzrv:
Looking at the remaining UGens, I realize, that there are two more dependants of `nova::`: `source/DWGUGens/dwglib/DWG.cpp` and `source/VBAPUGens/VBAP.cpp`. Those compile without additional specific use of the nova-tt source. Can we therefore assume, that exposing the SuperCollider libs is enough? I will try this later with NovaDiskIO, as this would mean to remove that check altogether (IMO).
Additionally there's an early dead-code copy of the NovaDiskIO source in `source/NovaUGens/`, that should be removed.

brianlheim:
> Looking at the remaining UGens, I realize, that there are two more dependants of nova::: source/DWGUGens/dwglib/DWG.cpp and source/VBAPUGens/VBAP.cpp.

DWG.cpp has that code commented out (it should be removed)

VBAP.cpp has nova code compiled conditionally on NOVA_SIMD, which only needs what's already present in `external_libraries` in this repo

Actually on closer inspection the plugins in `NovaDiskIO` don't need nova-tt at all, and the plugins in `NovaUGens` are never compiled, so you're right that we can remove both `NovaUGens` and the check for nova-tt entirely.

Going back a few messages:

> As pointed out before: SC_PATH both meaning SuperCollider source and compiled version location is a pretty unfortunate state.

> This is the only Ugen, that really requires the source code of SuperCollider though (because of nova-tt).

I don't understand what either of these statements mean. Can you please elaborate? Keep in mind that on platforms other than Linux, SuperCollider headers are not installed by the install target.

brianlheim:
In the interest of moving forward, could we perhaps merge this as-is (as it does solve a known issue) and discuss what to do about NovaDiskIO on a new ticket? I think there is a lot more to discuss and I don't want to do that here unless it's absolutely relevant.

dvzrv:
> DWG.cpp has that code commented out

... and that's why I hate block comments :>

> I don't understand what either of these statements mean. Can you please elaborate? Keep in mind that on platforms other than Linux, SuperCollider headers are not installed by the install target.

Oh, I didn't know that! Please forget what I wrote then... ;-)
Thinking about it, it should probably be clarified, what it can't point at on what OS.

> In the interest of moving forward, could we perhaps merge this as-is (as it does solve a known issue) and discuss what to do about NovaDiskIO on a new ticket? I think there is a lot more to discuss and I don't want to do that here unless it's absolutely relevant.

Yes, please let's do it. AFAICT NovaDiskIO doesn't compile due to boost incompatibilities right now anyways, but that really needs to go into a separate topic!

CMakeLists.txt
```diff
@@ -36,6 +40,19 @@ else()
...
 
+set(NOVA_TT_MISSING_ERROR "The nova-tt source code is missing in \
+${SC_PATH}/external_libraries/nova-tt.\n Make sure to point to a valid version \
+of SuperCollider's source code (with the help of the SC_PATH variable).\n It's \
+easiest to use a release tarball of SuperCollider.\n If you use a git clone, \
+make sure you have all submodules by running `git submodule update --init \
+--recursive` in the root of SuperCollider's source code repository.")
+
+if (NOVA_DISK_IO)
+  if (NOT EXISTS "${SC_PATH}/external_libraries/nova-tt/CMakeLists.txt")
```
dvzrv: This needs to be revised and then potentially changed/removed in another pull request.

brianlheim:
> Thinking about it, it should probably be clarified, what it can't point at on what OS.

I guess because it's always been this way for me, I just assumed "point to sc source" means "point to a git clone of the sc repo". but of course that is not totally clear when source headers can also be somewhere else.

brianlheim:
I opened #239, #240, #241 for the topics brought up in this PR. Please leave future comments on one of those (or open a new ticket if needed). Thanks again for your patience @dvzrv

PR #230, opened 2018-12-03
Update DWGPluckedStiff.schelp

redFrik:
fixed missing release argument. as reported on sc-users https://www.listarc.bham.ac.uk/lists/sc-users/msg62915.html


brianlheim:
Thanks!

PR #231, opened 2018-12-05
Travis: +supernova, other small improvements

brianlheim:
- build supernova plugins on linux and macOS (they will be included in macOS release packages now)
- shallow clone supercollider
- build with -j to get max cores on both linux and macOS


brianlheim:
will rebase this onto 3.10 once #233 is merged

PR #238, opened 2018-12-16
New plugin: BFold (was: Timbre) (wavefolder) 

georgezachos:
Virtual analog model of the "timbre" wavefolding circuit from the Buchla 259 Complex Waveform Generator.

### DIFF
--- a/source/VAfxUGens/BFold.cpp
+++ b/source/VAfxUGens/BFold.cpp
@@ -84,7 +84,7 @@ private:
 };
 
 // the entry point is called by the host when the plug-in is loaded
-PluginLoad(TimbreUGens)
+PluginLoad(VAfxUGens)
 {
     // InterfaceTable *inTable implicitly given as argument to the load function
     ft = inTable; // store pointer to InterfaceTable


snappizz:
nice work! i'll give this a closer look later. i'm wondering if this could have a more specific name, like `BFold` or `B259Fold`?

georgezachos:
thanks! yes, either of those names are better actually, less generic. `BFold` might be fine, I'm not aware of other buchla popular wavefolders.

source/VAfxUGens/sc/HelpSource/Classes/BFold.schelp
```diff
@@ -0,0 +1,29 @@
...
+Folding amount (audio rate). Useful range: [0, ..]
+
+argument::offset
+Input offset (control rate). Useful range: [-1, 1].
+
+
+Examples::
+
+code::
+{Timbre.ar(SinOsc.ar(200), MouseX.kr(0,3), MouseY.kr(-1,1))}.scope
```
patrickdupuis: The UGen is no longer called `Timber`. Also, according to the documentation above, `fold` needs to be audio rate. You might want to do `K2A.ar(MouseX.kr(0, 3))` instead.

source/VAfxUGens/sc/HelpSource/Classes/BFold.schelp
```diff
@@ -0,0 +1,29 @@
+class:: Timbre
+summary:: Wavefolder virtual analog model
+related:: 
+categories:: UGens>Filters
+
```
patrickdupuis: I'm not certain what the problem is exactly, but there's a syntax error in this help file. I get this message in the post window:

```
ERROR: In /usr/local/share/SuperCollider/Extensions/SC3plugins/VAfxUGens/HelpSource/Classes/BFold.schelp:
  At line 4: syntax error, unexpected newline, expecting text or URL

```

source/VAfxUGens/sc/HelpSource/Classes/BFold.schelp
georgezachos: Oops, rookie mistakes, sorry. Fixed these now.

source/VAfxUGens/sc/HelpSource/Classes/BFold.schelp
georgezachos: I think it was the empty `related::` field. I removed it (for now)

source/VAfxUGens/sc/BFold.sc
```diff
@@ -0,0 +1,5 @@
+BFold : UGen {
+    *ar { arg in, fold=0.0, offset=0.0;
+        ^this.multiNew('audio', in, fold, offset)
+    }
```
patrickdupuis: Since the first two arguments have to be audio rate, it would be a good idea to check the input rates. I think this is how it's done (taken from example plugins repo https://github.com/supercollider/example-plugins/blob/master/01a-BoringMixer/BoringMixer.sc)

```supercollider
	checkInputs {
		[0, 1].do { |i|
			if (inputs[i].rate != \audio) {
				^"input % is not audio rate".format(i).throw
			}
		};
		^this.checkValidInputs;
	}
```

@snappizz is this correct?

georgezachos: Ah yes! For some reason the BoringMixer's string format inside the `if` statement didn't work, but the one used in bbc6a908bc8741e82cd02dd5df91b38ed35bce48 is ok.

georgezachos:
One last fix with the 388a90392e5b2bab01e154d122d6966d904d13a5. Sorry for the sloppiness, will be more careful when I PR again.

brianlheim:
@georgezachos Can you please update this PR per @snappizz 's request? Thanks!

georgezachos:
Sorry for the delay @brianlheim and thanks for the patience! I updated the PR now.

dyfer:
@yorgoszachos sorry this hasn't been attended to. Could you please rebase (again...) on to the current `main`?
@patrickdupuis do you think this PR is good to go, aside from the rebase?

capital-G:
Hey @georgezachos - would still be very cool to get these in here. Could you rebase against main so that we can have a CI run of this? If you need some help, hit me up.

PR #251, opened 2019-02-09
Fix Travis build on Linux

snappizz:


brianlheim:
whoops, my bad. changing this back to -j2 should be fine.

snappizz:
lol. ok, good to go now -- github marks the CI status as "pending" due to use of [skip appveyor], but this should be fine

PR #255, opened 2019-03-09
Update README to reflect frozen status

snappizz:
see discussion at https://www.listarc.bham.ac.uk/lists/sc-dev/msg58832.html for background.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,10 @@
 # sc3-plugins
 
+sc3-plugins is a collection of unit generator plugins for [SuperCollider](http://supercollider.github.io/) contributed by users between 2006 and 2018. (Plugins extend the functionality of the SuperCollider server, and are distinct from [quarks](https://github.com/supercollider-quarks/), which are packages for the SuperCollider programming language.)
 
-This repository contains the community collection of unit generator plugins for [SuperCollider](http://supercollider.github.io/).
-An installation extends the functionality of SuperCollider by additional UGens that run on `scsynth`, the SuperCollider audio synthesis server.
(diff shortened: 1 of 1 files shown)


LFSaw:
Interesting idea, although IMHO this PR seems to be a little too far ahead of discussion. However, since discussion on sc-dev seems to have stalled, I'll try to open it up again here.

I put my thoughts about this into this email to sc-dev: https://www.listarc.bham.ac.uk/lists/sc-dev/msg58854.html

I can condense this down into several questions:

+ When using the cookiecutter plugin develop environment, how do I ensure my plugins to be seen/browsed by the community?
+ How Do I move my existing plugins out of sc3-plugins without breaking people's workflow (duplicate plugins installed etc.)?
+ How do I easily install/deinstall plugins not in this repository?

patrickdupuis:
@LFSaw the conversation has indeed stalled. Would you mind re-igniting it with the questions you've posted above?

LFSaw:
> Would you mind re-igniting it with the questions you've posted above?

done. :)

mtmccrea:
I agree this is a bit ahead of discussion and the proposal needs to have, at a minimum a documented "preferred" way to authors to produce UGens ready for install from their own repos (i.e. a release procedure, not just how to make a UGen), and for users to download and install UGens from out in the wild.

telephon:
>When using the cookiecutter plugin develop environment, how do I ensure my plugins to be seen/browsed by the community?

I suppose that UGens would be listed as part of the quarks directory file, wouldn't they?

>How Do I move my existing plugins out of sc3-plugins without breaking people's workflow (duplicate plugins installed etc.)?

This is unsolved so far.

LFSaw:
fyi I'll refrain from taking part in this discussion. Reasons here:

https://github.com/supercollider/incubator/issues/1

PR #260, opened 2019-06-05
Make LadspaUGen optional

### LINKED ISSUE #259: Cmake option to disable LadspaUGen
LadspaUGen is known to prevent the server from booting in some Linux environments. See #257 (erroneously closed because of a misdiagnosis -- continued at https://github.com/supercollider/supercollider/issues/4421 where it turned out to be an sc3-plugins issue after all).

LadspaUGen hasn't been maintained in several years. See #23 (4.5 years old, unfixed) and #210 (9/2018).

If it isn't going to be fixed (and I have no confidence that it will be), then we should provide a way for users to turn it off.

gusano:
Fixes #259

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,7 @@ else()
 endif()
 
 option(AY "Build with AY ugens" ON)
+option(LADSPA "Build with Ladspa ugen" ON)
 option(QUARKS "Install plugins as quarks")
 option(OSX_PACKAGE "Package dmg for Apple")
 option(IN_PLACE_BUILD "Build and install in cmake build folder" ON)
--- a/README.md
+++ b/README.md
@@ -86,10 +86,10 @@ ambisonics) subproject.
     * (macOS ONLY, default=ON)
     * `cmake -DIN_PLACE_BUILD=ON`
 + Build the plugins as quarks
-    * (default 'OFF')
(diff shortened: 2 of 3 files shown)


patrickdupuis:
I get this error when trying to run `cmake -DLADSPA=OFF ..`

```
CMake Error at source/CMakeLists.txt:363 (install):
  install TARGETS given target "ladspalist" which does not exist in this
  directory.
```

gusano:
@patrickdupuis doh.. I fixed it and force pushed.. sorry for that and thanks for spotting

patrickdupuis:
Thanks! Tested on Linux.

brianlheim:
There should be more documentation than this -- maybe a known issues section in the readme?

gusano:
@brianlheim note added in the readme

totalgee:
This frequent source of confusion ("why does scsynth fail to boot on my Linux distro?") came up again in the TidalCycles forums recently:
https://club.tidalcycles.org/t/supercollider-wont-boot/2285

The "solution" of deleting the LADSPA plugin locally was discussed, and I sent a link to this PR. The comment was made that hopefully it would be merged soon... ;-)

brianlheim (approved):

totalgee:
Thanks @brianlheim !

PR #261, opened 2019-06-13
increase kMaxSynthGrains in JoshGrainUGens

dyfer:
Occasionally I am running into exceeding the number of grains.
This just arbitrarily exceeds to maximum, but maybe there is a better way to do this?

### DIFF
--- a/source/JoshUGens/JoshGrainUGens.cpp
+++ b/source/JoshUGens/JoshGrainUGens.cpp
@@ -87,7 +87,7 @@ struct WinGrainI
 
 const int kMaxGrains = 64;
 
-const int kMaxSynthGrains = 512;
+const int kMaxSynthGrains = 2048;
 
 struct MonoGrain : public Unit
 {


joshpar:
thanks @dyfer - there is, actually, and we do it in the main line grain UGens that were based on these UGens. Basically, you add a parameter in the UGen that lets the user pass in the max number of grains, and RTAlloc the space for them. That is a big change to THESE UGens though and may break backward compatibility. You can see an example from this page: https://github.com/supercollider/supercollider/blob/develop/server/plugins/GrainUGens.cpp
`void GrainFM_next_k(GrainFM* unit, int inNumSamples) {
    ClearUnitOutputs(unit, inNumSamples);

    if (unit->mFirst) {
        unit->mFirst = false;
        float maxGrains = IN0(7);
        unit->mMaxGrains = (int)maxGrains;
        unit->mGrains = (GrainFMG*)RTAlloc(unit->mWorld, unit->mMaxGrains * sizeof(GrainFMG));
    }

    GrainFM_next_play_active(unit, inNumSamples);

    float trig = IN0(0);
    if ((unit->curtrig <= 0) && (trig > 0.0))
        GrainFM_next_start_new<false>(unit, inNumSamples, 0);

    unit->curtrig = trig;
}

void GrainFM_Ctor(GrainFM* unit) {
    if (INRATE(0) == calc_FullRate)
        SETCALC(GrainFM_next_a);
    else
        SETCALC(GrainFM_next_k);
    int tableSizeSin = ft->mSineSize;
    unit->m_lomask = (tableSizeSin - 1) << 3;
    unit->m_radtoinc = tableSizeSin * (rtwopi * 65536.);
    unit->m_cpstoinc = tableSizeSin * SAMPLEDUR * 65536.;
    unit->curtrig = 0.f;
    unit->mNumActive = 0;
    unit->mFirst = true;
    GrainFM_next_k(unit, 1);
}`
Notice the RTAlloc in next_k IF this is the first run. The value is read from the UGen input, then allocated. I'd suggest first seeing if something similar can be done without breaking backward compatibility, and if so, that would be a better fix. If backward compatibility would be broken, I'm fine with what you have already submitted.... please let me know!

dyfer:
Thanks @joshpar I'll look into it.
It was a change I made a while back to fix the number of grains. I'll try to look at the "proper" fix later and will update this PR then.

capital-G:
bump @dyfer

PR #262, opened 2019-06-18
JoshPVUGens memory allocation fix

Spacechild1:
use _alloca instead of std::vector for MSVC to avoid calling malloc on the RT thread.

Not tested! It should work fine, but I don't have the time to compile/test on Windows.

EDIT: as @joshpar pointed out, we could also just use RTAlloc instead of stack allocation.

### DIFF
--- a/source/JoshUGens/JoshPVUGens.cpp
+++ b/source/JoshUGens/JoshPVUGens.cpp
@@ -10,6 +10,9 @@
 //third party Phase Vocoder UGens
 #include <vector>
 #include <algorithm>
+#ifdef _MSC_VER
+# include <malloc.h> // for _alloca()
+#endif
 #include "SC_fftlib.h"
 #include "FFT_UGens.h"
 #include "SCComplex.h"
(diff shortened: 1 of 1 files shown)


joshpar:
@Spacechild1 - I can't remember... did we discuss this recently? While looking at the change, I actually think the solution is to use the RTAlloc / real-time memory pool. The original code (before it was wrapped up for Windows) appears to be naive c++ programming on my part. I'll try to work on this soon. But - before I do, any concerns on your end with me going that route for refactoring this?

Spacechild1:
>  can't remember... did we discuss this recently? 

Yes, we talked via e-mail. 

> I actually think the solution is to use the RTAlloc / real-time memory pool.

My alloca() solution just mimicks the behavior of variable length arrays (we use alloca() a lot in Pd, btw). Of course, you can just use RTAlloc on all platforms instead. Feel free to make another PR and close this one!

joshpar:
Thanks! I'll put together a pull request this weekend and ask you to test if that's ok - just to make sure things work as expected across platforms.

PR #273, opened 2020-02-05
better vocoder

bramtayl:
I based this on [Audacity's implementation](https://github.com/audacity/audacity/blob/master/plug-ins/vocoder.ny). Not sure what that implies for licensing

### DIFF
--- a/source/JoshUGens/sc/classes/Vocoder.sc
+++ b/source/JoshUGens/sc/classes/Vocoder.sc
@@ -31,6 +31,7 @@ Vocoder {
 		^out + ( hpfscal * hf )}
 	// code based on the Audacity plug-in here:
 	// https://github.com/audacity/audacity/blob/master/plug-ins/vocoder.ny
+	// Credit to Audacity® and "Edgar-RFT"
 	*audacity {
 		arg carrier_wave, model_wave, number_of_bands,
 		lowest_band = 20,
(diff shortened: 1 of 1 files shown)


joshpar:
Thanks! Does this change the sound of the original class? If so, I'd prefer to have it us a different different *method name to retain the original. This isn't a judgement about whether it is better or not - but if it is different it represents a breaking change that users may not expect. You can see the *bark version as an example. As far as licensing, it is a concern if the algorithm or any of the parts of the recipe are owned by someone else. It may be a question for someone on the Audacity team OR it may fit in just fine with a GPL license.

bramtayl:
Sure. What should I call the new method? In the audacity code the header says "Released under terms of the GNU General Public License version 2" so I'm guessing there shouldn't be too many problems?

joshpar:
You can call it *audacity if you want - I'd also include (in the file) a comment that describes where you got it, and links to any sources you referenced. IF there is an author mentioned for it on the Audacity page, I'd suggest reaching out to them to ask permission to use it or to at least let them know (as a courtesy). I've done this before as well.

SteveDaulton:
The vocoder effect in Audacity is not a "traditional" vocoder. It's a quirky kind of "vocoder-like" effect. As such, I doubt that it would be a suitable replacement for SuperCollider's vocoder, though it could be considered as an additional effect.
As with the main Audacity application, the effect is licensed under the terms of GPL v2.

If you do use it, then it would be nice (though not compulsory) for you to credit Audacity® and "Edgar-RFT" (the original developer of this effect), possibly as a code comment.

bramtayl:
Sure. I have no understanding of how these things work, but at least to my ears the vocoder from Audacity sounds a lot better from the vocoder from supercollider. The implementations seem fairly similar, so I suspect that whatever quirkiness applies to the Audacity implementation applies to the supercollider one as well.

jreus:
Did this ever get ported over? I'd love to have more options for vocoder implementations in SC. The built-in Vocoder Ugen is very difficult to get a musical sound with.

> Sure. I have no understanding of how these things work, but at least to my ears the vocoder from Audacity sounds a lot better from the vocoder from supercollider. The implementations seem fairly similar, so I suspect that whatever quirkiness applies to the Audacity implementation applies to the supercollider one as well.

dyfer:
I don't think this process was ever finished.
sc3-plugins are in a transitory state where we as a community are trying to decide how to best go about its future, with no clear path forward yet. While no "official" decision has been made, we might not be accepting new additions to the sc3-plugins while we are in this transitory state.

PR #280, opened 2020-04-07
enable supernova plugins build on travis

### LINKED ISSUE #264: macOS release doesn't include supernova plugins
Now that supernova is provided in official macOS builds, I think it would be reasonable to also include plugins for supernova on macOS in the release download.

dyfer:
Fixes #264


brianlheim:
superceded in #281. i knew we had fixed this before...

PR #281, opened 2020-04-07
Merge 3.10 into master

brianlheim:
I merged 3.10 into this branch locally and resolved conflicts. The major conflicts were in travis.yml, and I preferred the 3.10 version although there were not many changes. There was also a deleted file in master that wasn't deleted in 3.10; I preferred deletion.

This supercedes #280.


brianlheim:
passed except for artifact packaging on appveyor -- we should consider using the same fix from main sc repo

PR #282, opened 2020-04-08
stop cmake warnings, issue #191

sonoro1234:
substitution of #196

But it seems that the suggested change broke compilation:

````
include could not find load file:
    ../../supercollider/SCVersion.txt

````
but it shouldnt because this file exists...

### CODE REVIEW

**Comment by dyfer** [comment]
thanks @sonoro1234 

For me this worked when I put
```
include("${SC_PATH}/SCVersion.txt")
set(PROJECT_VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}${PROJECT_VERSION_PATCH}")
message(STATUS "Building plugins for SuperCollider version: ${PROJECT_VERSION}")
```
between
```cmake_policy(SET CMP0048 NEW)```
and
```project (sc3-plugins VERSION ${PROJECT_VERSION})```

I also noticed that github still indicates that this branch is out-of-date. Have you updated your master before creating this branch?

**Comment by sonoro1234** [comment]
>  Have you updated your master before creating this branch?

Yes but it seems to  run so quickly lately...  Thats good!!!
I will remake it tommorow.

Thanks.

File: CMakeLists.txt [hunk 1]
```diff
@@ -1,6 +1,10 @@
-cmake_minimum_required (VERSION 2.8)
-set(SUPERNOVA_CMAKE_MINVERSION 3.1)
-project (sc3-plugins)
+include("${SC_PATH}/SCVersion.txt")
```
brianlheim, line 1: @dyfer looking at this more closely now. this requires SC_PATH to be set explicitly by the user whereas before find_package(SuperCollider3) would set it automatically. I think that's fine, but then build instructions should be reviewed and `cmake_modules/`FindSuperCollider3.cmake` and its reference in CMakeLists.txt should be removed. if that doesn't sound good then i think we need to go back to something like it was before, because AFAICT calling find_package() before cmake_minimum_required and project() is not allowed

sonoro1234:
I can build locally against supercollider/develop.
I dont know what happens with travis and appveyor

brianlheim:
Appveyor and Travis are correct; you will need to get both to pass with the way they use SC_PATH. I'm sorry, but I really don't have much desire to try to fix this since it is a very small issue that has taken up more time than I would like.. maybe you can do some research, and ping me again when you think it's ready to merge.

sonoro1234:
>  I'm sorry, but I really don't have much desire to try to fix this since it is a very small issue that has taken up more time than I would like..

I also wasnt interested but https://github.com/supercollider/sc3-plugins/issues/191#issuecomment-375105061

brianlheim:
ok! thanks for giving it a shot

PR #283, opened 2020-04-08
appveyor: turn off artifacts, push to s3 directly

brianlheim:


brianlheim:
counterpart to https://github.com/supercollider/supercollider/pull/4861

brianlheim:
this is now working as best i can tell! i just forced push to consolidate my work.

PR #284, opened 2020-04-10
Topic/fix bhob grain ugens

scztt:
This fixes three major functional issues with the TGrains ugens, and one major code issue.

FIXED: TGrains2/3 did not work with an audio rate trigger. Both UGens are set to support aliasing input and output buffers. However, they processed grain output FIRST, overwriting the output buffer in the process. If this buffer was aliased, then this write overwrote the input triggers, causing the synth to fail to properly trigger. The solution is to process triggers FIRST, and then process output of running grains.

FIXED: Because triggers were processed after grain output, new grains were not being properly processed, resulting in a single buffer delay on grains. On code inspection, it appeared to me that this was written correctly - but after rearranging for the previous fix, the new output started at the correct time for when it was triggered, whereas the old UGen implementation showed a delay.

FIXED: TGrains2/3 did not support local buffers. This was the result of old-style buffer lookup.

FIXED: The implementations of TGrains2/3 are almost identical, with only small differences in inner loop details and variable lookup. This change refactors both into a single templatized function, with a few constexpr branches to specify specific behavior for the two UGens.

Note that, while I did do some minor refactoring in this PR, it is NOT an attempt to fully clean up these UGens. Lots of archaic patterns and non-ideal code was left as-is because, frankly, this kind of grain processing is extremely sensitive, and the risk of breakage is quite high. 

All changes in this PR were A/B tested against the original output of TGrains2/3, and have differences of much less than -48db. There *IS AN OUTPUT CHANGE* in that the output is no longer delayed by one block size (I accounted for this when comparison testing) - however, this is too large a bug to leave the former, incorrect behavior in place. There is currently no facility for pin testing / unit testing of UGen output, but the script I used is here for reference: https://gist.github.com/scztt/75c042ea0ce6b184cceb900781020bb0

### DIFF
--- a/source/BhobUGens/BhobChaos.cpp
+++ b/source/BhobUGens/BhobChaos.cpp
@@ -5,7 +5,7 @@ http://www.bhobrainey.net
 
 SuperCollider real time audio synthesis system
     Copyright (c) 2002 James McCartney. All rights reserved.
-	http://www.audiosynth.com
+    http://www.audiosynth.com
 
     This program is free software; you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


scztt:
Fixes #284 272

brianlheim:
i'll take your word that it performs similarly to before. build is failing because you're using a lot of c++14 and 17 features (`constexpr` extended support, `constexpr if`)

brianlheim:
umm.. i don't think requiring C++17 is a good idea.. CI and documentation have to be updated then, and it will break lots of builds; it shouldn't be too bad to do it in C++11 right? aside from `hasWindowBuffer` and `interpolationArgIndex` i don't think the other uses are even helping. the `if constexpr` block will most certainly be optimized out without the `constexpr` anyway, and `findBuf` is never used in a place where it can be constexpr-evaluated.

scztt:
Looks like our linux travis config is just too far behind to support c++17, sadly. It's probably possible to implement these without `constexpr` and still get the compiler to optimized correctly. I'll probably pull the refactoring change out, so this PR contains only the bug fixes. It still leaves the code in an ugly state and is probably a little slower, but I don't want to hold up these fixes as a lot of people have been asking and I don't know when I'd be able to re-write this with c++11 (or if it's worth it, rather than waiting until travis dumps trusty and we have to upgrade :)).

lnihlen:
FWIW, Travis supports Linux build environments as new as Bionic:
https://docs.travis-ci.com/user/reference/overview/#linux. I'm building
Scintillator against Bionic, as I also use extensive C++17 stuff and didn't
want to faff around with compatability libraries.

brianlheim:
i think the changes i noted above should be enough to make this C++11 compatible. i actually don't think `constexpr` is helping you much here at all.

https://github.com/supercollider/rfcs/pull/1 is relevant; C++17 support is probably not too far around the corner but i don't think it's a good idea to force this on people without warning.

lnihlen:
That's an excellent point, and the RFC is absolutely the correct forum for
further discussion about this matter. It would be poor leadership indeed to
decide to move to C++17 in the middle of an only-tangentally-related PR. I
wanted only to add to the record that the C++17 support in Travis is there.

scztt:
The `constexpr if` protects access to members that don't exist in all cases - the workaround there is to break that out to a separate templatized function. The others are likely to be optimized with just a const vs constexpr, but I wouldn't push that without a little more research to prove that. In the end, that was only a code quality change for the next person that comes along - I pushed a branch with the single-implementation separately, it probably should not be mixed in to a bug-fix PR anyway.

scztt:
As an aside: in researching techniques for consolidating different inner loop implementations into a single piece of code, I did a fair amount of testing to see how far it was possible to push compilers in this area.... I was pretty delighted to find that I was able to do a few layers of function composition of lambdas (e.g. capturing or passing conditionally defined constexpr lambdas as arguments to other constexpr lambdas) inside of a loop and the latest clang was still able to optimize the entire thing down to an integer assignment of the result value.

brianlheim:
> The others are likely to be optimized with just a const vs constexpr, but I wouldn't push that without a little more research to prove that.

sorry, i don't understand... are you referring to the lambdas? are compilers more likely to inline the call operators of lambdas in constexpr variables over non-constexpr? i thought that just means the lambda initialization _can_ be compile-time evaluated.

scztt:
Mainly, I just meant: I would want to check the code gen before committing a non-constexpr version of this code, to make sure its still generating what I expect. Afaik there shouldn't be any difference between the implementations you're thinking of. I'm being probably more pedantic than necessary, but I've been deep in microbenchmarking TGrainsN to squeeze a little more performance from it, so my inclination is to assume any changes are bad until I see some numbers go down :)

brianlheim:
alright, makes sense!

scztt:
It looks like Appveyor is hung on on the branch build - I tried rebuilding once, but that didn't resolve...

brianlheim:
the branch build should succeed if you rebase on master (after #283 ). but it's not particularly important for that one to pass anyway.

brianlheim:
@scztt the sc3plugins release for 3.11 hasn't been made yet. are you aiming to get this change in there?

scztt:
I don't have a strong preference. There have been several mailing list / forum questions lately about TGrains2/3 not working - it's entirely broken for audio rate arguments and LocalBuf, which is pretty severe, so I'd say it's definitely worth considering.

brianlheim:
discussed this on the call today and Josh, me, and a few other people all thought this is fine to include in 3.11, so let's do it :) thanks!

PR #207, opened 2018-07-31
added lightweight encoders 

florian-grond:
In this PR Thibaud Keller suggested to add some lightweight encoders as additional options for encoding HOA sources, also there are max re and in phase filters for the b-format up to order 5

PR #291, opened 2020-06-04
additional lightweight HOAUgens

thibaudk:
Hello,

this PR is supposed to solve the issues of #207.
After a few back and forth, I do think that these additions have their interest,
and the build time is not that much longer than with the HOAUgens at the moment.
I may look into a jumble build next, hoping to reduce the number of targets...

tested on Ubuntu 20.04
and supercollider 3.11

### DIFF
--- a/source/HOAUGens/HOALibEnc3D1.cpp
+++ b/source/HOAUGens/HOALibEnc3D1.cpp
@@ -1,0 +1,1192 @@
+/* ------------------------------------------------------------
+author: "Pierre Guillot, Eliott Paris"
+copyright: "2012-2015 Guillot, Paris, CICM, Labex Arts H2H, U. Paris 8"
+name: "HOALibEnc3D1"
+Code generated with Faust 2.18.0 (https://faust.grame.fr)
+Compilation options: -lang cpp -scal -ftz 0
+------------------------------------------------------------ */
+
+#ifndef  __mydsp_H__
+#define  __mydsp_H__
+
+/************************************************************************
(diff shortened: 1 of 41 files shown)


dyfer:
Hello @thibaudk 
Thank you for your PR and sorry for not attending to this earlier.
That's quite a number of new UGens! From previous discussions on HOAUgens, it seemed that the general direction was to possibly move them all outside of sc3-plugins, rather than adding more to them to this repository. How do you feel about this addition @telephon @patrickdupuis @joshpar ?

thibaudk:
Hi @dyfer,

Actually, The HOAUgens are largely based on @sekisushai 's [ambitools](https://github.com/sekisushai/ambitools) that was recently relocated and updated. The new dsp's can be found [here](http://sekisushai.net/git/sekisushai/ambitools/src/branch/master/faust).

I should look into re-basing HOAUgens and @florian-grond 's [SC-HOA](https://github.com/florian-grond/SC-HOA) on this new version that seem to amount to less plugins.

thibaudk:
Hi, very sorry, I have not touched it since. So I suppose you can consider the PR stale at this point.
The files were generated with Faust 2.18.0 against the SuperCollider version at the time. The current Faust is 2.74.6, so it may be a little outdated.

dyfer:
I feel that we shouldn't be pulling in that many plugins into sc3-plugins - it should be its own plugin suite. As this is stale, I suggest closing this.

capital-G:
Then let's close it - reopen if I necessary.

PR #293, opened 2020-06-05
added a note about using sudo under linux

redFrik:
and an extra 'cd sc3-plugins' to make sure users are in the right directory to begin with.
the sudo is needed to avoid the following error...
```
Install the project...
-- Install configuration: ""
-- Installing: /usr/local/share/SuperCollider/Extensions/SC3plugins/local
CMake Error at source/cmake_install.cmake:41 (file):
  file INSTALL cannot make directory
  "/usr/local/share/SuperCollider/Extensions/SC3plugins/local": No such file
  or directory
Call Stack (most recent call first):
  cmake_install.cmake:42 (include)


make: *** [Makefile:74: install
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,10 +58,11 @@ replace `/path/to/sc/` with the path to the SuperCollider source directory. That
 `../../supercollider` if you cloned both repositories in the same working directory.
 
 ```shell
+cd sc3-plugins
 mkdir build && cd build
 cmake -DSC_PATH=/path/to/sc/ ..
 cmake --build . --config Release
-# to install the plugins
+# to install the plugins - note: linux users likely need sudo
 cmake --build . --config Release --target install
 ```
 


dvzrv:
FWIW: Technically speaking `sudo` is not required, but running `make install` as `root`.
This can also be achieved using e.g. `su`.

That being said: Using `sudo make install` will only work if the given user (or any of the groups the user is attached to) is allowed to elevate its privileges to root (as defined in the `/etc/sudoers` configuration file or a dropin configuration file in `/etc/sudoers.d/`).

It would therefore be more correct to write:
*The following command needs to be run as root (e.g. using sudo or su).*

Btw: this is not only true for Linux, but also for e.g. macOS, as by default the `/usr/local` directory is not a "free for all" directory. However, the use of tools such as homebrew or macports change the permissions of that directory to be able to use it without elevating privileges all the time.

redFrik:
@dvzrv sure, but i wanted to keep it short and to the point.

and no, on macOS we don't by default need root (rather it should be discouraged).  see further down in the readme.  this is Linux only.

how about this then...
_# to install the plugins - note: For Linux the following command needs to be run as root (e.g. using sudo or su)._

which is way longer than i'd like it.

brianlheim (approved):
this phrasing is accurate enough for me, although the points @dvzrv raises are valid. i'm going to merge this now so it can be included in the the 3.11.1 release

PR #298, opened 2020-09-05
Update docs - fix cmake variable reference, VS versions

ethancrawford:
The plugins have successfully been compiled using VS 2019 (community edition) so
the compatible VS versions have been updated. It also required a change to one
of the cmake variable names.


brianlheim (approved):
thanks!

Issue #317, opened 2021-07-23
ATK: FOA speed of sound


joslloand:
ATK issue: https://github.com/ambisonictoolkit/atk-sc3/issues/104

> AtkFoa.speedOfSound = 333.0; // (m/s)
> 
> NOTE: this value is implied by the current setting of the cutoff
> frequency (53 Hz) found in the sc3-plugins code for FoaProximity & FoaNFC
> 
> freq = speedOfSound / 2pi // wavenumber = 1
> speedOfSound = 53 * 2pi // 333.0
> 
> this value is the speed of sound @ 3C
> https://www.weather.gov/epz/wxcalc_speedofsound
> 
> TODO: consider updating sc3-plugins to align with
> AtkHoa.speedOfSound = 343.0

The fix for this is simple. For FoaProximity and FoaNFC, replace the frequency setting of 53.0 Hz with 343.0 / 2pi -> 54.59014548052

There are four lines to be updated:
https://github.com/supercollider/sc3-plugins/blob/fa550fbbe28f78ac1872ed580b21b78ef6b63c42/source/ATK/AtkUGens.cpp#L1949
https://github.com/supercollider/sc3-plugins/blob/fa550fbbe28f78ac1872ed580b21b78ef6b63c42/source/ATK/AtkUGens.cpp#L1999
https://github.com/supercollider/sc3-plugins/blob/fa550fbbe28f78ac1872ed580b21b78ef6b63c42/source/ATK/AtkUGens.cpp#L2063
https://github.com/supercollider/sc3-plugins/blob/fa550fbbe28f78ac1872ed580b21b78ef6b63c42/source/ATK/AtkUGens.cpp#L2113

joslloand:
Don't forget to update the [README](https://github.com/supercollider/sc3-plugins/blob/main/source/ATK/sc/README.md)!

joslloand:
@mossheim, @telephon, @joshpar 

The question is one of etiquette. Is it preferred for me to directly make the above listed changes to the ATK's plugins, or would it be preferable for me to make a PR?

Do let me know, as I'd like to be sure play nicely!

dyfer:
Hi @joslloand 
I think anything that gets into the main branch needs to go through a PR anyway, so making one is probably the best way forward.

joslloand:
Thanks @dyfer, I'll go ahead and prepare a PR.

It would be great to be able to have the update in to coincide with the upcoming SC 3.12 release.

[...]

PR #319, opened 2021-07-27
Topic/atk fix speed of sound

joslloand:
PR to address #317 (& ATK Quark [104](https://github.com/ambisonictoolkit/atk-sc3/issues/104))

Additional updates to [ATK sc3-plugins README](https://github.com/supercollider/sc3-plugins/blob/main/source/ATK/sc/README.md) to align with [ATK Quark README](https://github.com/ambisonictoolkit/atk-sc3/blob/master/README.md) & in preparation for sc3-plugins v3.12.0 release.


dyfer:
Hi @joslloand 
I haven't looked at this closely, but wouldn't it be possible to have this value set from sclang?

source/ATK/AtkUGens.cpp
```diff
@@ -1944,9 +1944,10 @@ void FoaNFC_next_k(FoaNFC *unit, int inNumSamples)
     float y1x = unit->m_y1x;
     float y1y = unit->m_y1y;
     float y1z = unit->m_y1z;
-    
+
     for(int i = 0; i < inNumSamples; i++){
-		float freq = 53.0 / distanceStart;
+		// AtkFoa.speedOfSound = 343.0  --> 343.0 / 2pi --> 54.59014548052
+		float freq = 54.59014548052 / distanceStart;
```
joshpar, line 1950: I think this is a good opportunity to store this value into a C constant, and that constant is used instead of the 54.590 ... etc.

joshpar (changes requested):

source/ATK/AtkUGens.cpp
joslloand: Good idea!

dyfer: Sorry if this is out of scope, but wouldn't we want to be able to change speed of sound? It depends on environmental factors, after all...

joslloand:
> I haven't looked at this closely, but wouldn't it be possible to have this value set from sclang?
@dyfer, ah, that might be better, as it would then be changeable?

Ah, this is a nice idea!

In practice, I don't think I'd want this to be an argument that the user supplies directly in the user interface to the UGen. Doing so could / would break existing code.

Instead, it should happen "under the hood" by setting `AtkFoa.speedOfSound` to the relevant value.

---
Ok, having had a little look, the changes to do so would have to happen in both [AtkUGens.cpp](https://github.com/supercollider/sc3-plugins/blob/main/source/ATK/AtkUGens.cpp) here in sc3-plugins, and in [ATK.sc](https://github.com/ambisonictoolkit/atk-sc3/blob/dev/Classes/ATK.sc#L950-L973) found in the quark.

@joshpar, what do you think of this?

dyfer:
> Instead, it should happen "under the hood" by setting `AtkFoa.speedOfSound` to the relevant value.

Yes, that's what I was thinking

joslloand:
> > Instead, it should happen "under the hood" by setting `AtkFoa.speedOfSound` to the relevant value.
> 
> Yes, that's what I was thinking

Good to know!

The _slight_ danger is that users will need to _make sure_ to update both sc3-plugins and the atk-sc3 quark. That shouldn't be a problem if the change is communicated in the update announcement..

@joshpar, how does this sound to you. (I'll add @mtmccrea, too.)

joshpar:
yeah @joslloand - I think that's a cool idea

joslloand:
> yeah @joslloand - I think that's a cool idea

Ok, great!

As for actually making it happen, I'm _super rusty_ on the plugin side and would need to have some guidance on getting it right.

In particular, once the speed of sound is set, it _should_ be constant for the life of the UGen instantiation. I.e., it _should_ be a fixed value that isn't modulate-able.

@dyfer, I may reach out over email for a quick chat on the matter....

joshpar:
I can help as well if needed. Cheers!

joslloand:
> I can help as well if needed. Cheers!

Ok, great! I think I see what needs to happen... so that's fine. (However, since, I haven't tried to compile sc3-plugins on this machine in so long, I'm expecting a task!)

source/ATK/AtkUGens.cpp
joslloand: Agreed... my latest push could be the thing...

joslloand:
Ok, I've just made an updated push.

The updated atk-sc3 quark branch is here: https://github.com/ambisonictoolkit/atk-sc3/tree/refactor-foa-nfc-proximity

What I haven't done is tried to compile... so testing is required....

dyfer:
@joslloand we've just added automated building with github actions. If you rebase your PR on the current `main`, you'll see an automated build in this PR, where you can also download the build artifacts to test. However, building locally is definitely recommended :)

source/ATK/AtkUGens.cpp
```diff
@@ -1937,20 +1938,17 @@ void FoaNFC_next_k(FoaNFC *unit, int inNumSamples)
...
     float y1y = unit->m_y1y;
     float y1z = unit->m_y1z;
 
     for(int i = 0; i < inNumSamples; i++){
-		// AtkFoa.speedOfSound = 343.0  --> 343.0 / 2pi --> 54.59014548052
-		float freq = 54.59014548052 / distanceStart;
-		float wc = (twopi * freq) * SAMPLEDUR;
-
-		//	a0 = (1 + (wc.cos.neg * 2 + 2).sqrt).reciprocal;
+		float wc = (speedOfSound / distanceStart) * SAMPLEDUR;
```
mtmccrea, line 1951: A minor issue but there's the opportunity to factor out `speedOfSound * SAMPLEDUR` here, as well as to check whether `distanceStart` differs from the previous block, and if it doesn't then
```
float wc = (speedOfSound / distance[i]) * SAMPLEDUR;
float a0 = 1 / (sqrt((cos(wc) * -2) + 2) + 1);
```
need not be re-calculated. I don't have a strong feeling about this kind of change at the moment, because it's quite clean as is.

mtmccrea: Factoring out `speedOfSound * SAMPLEDUR` would be a matter of saving this value as a member var in the `Ctor`.

joslloand:
> @joslloand we've just added automated building with github actions. If you rebase your PR on the current `main`, you'll see an automated build in this PR, where you can also download the build artifacts to test. However, building locally is definitely recommended :)

Thanks @dyfer! I didn't realize this... very convenient. (I ended up doing a merge rather than a rebase.)

I've done some tests, and both HOA and FOA are now matched. Hurrah! So the updates are working now.

... so... I think we're ready for a merge

source/ATK/AtkUGens.cpp
joslloand: @mtmccrea, thanks for having a look here.

I'm not super keen as to tracking down optimizations with this update. We won't be running any slower than the previous incarnation, so it doesn't really feel especially worth it.

There is an argument to be made for an optimization sweep on the plugins further down the line, maybe....

mtmccrea: @joslloand I agree, we can revisit it if @dyfer wants to encode 500 nf sources in realtime and needs a little headroom ;)

joslloand:
> > @joslloand we've just added automated building with github actions. If you rebase your PR on the current `main`, you'll see an automated build in this PR, where you can also download the build artifacts to test. However, building locally is definitely recommended :)
> 
> Thanks @dyfer! I didn't realize this... very convenient. (I ended up doing a merge rather than a rebase.)
> 
> I've done some tests, and both HOA and FOA are now matched. Hurrah! So the updates are working now.
> 
> ... so... I think we're ready for a merge

@dyfer, @joshpar just picking this up again after some time away....

... I think this is all ready to go. The tests (matching NFE results w/ the HOA implementation) have passed.

So, I think this is ready to be closed and merged.

telephon:
@joshpar would you also agree that it is ready to go?

joslloand:
> @joshpar would you also agree that it is ready to go?

@telephon I had a conversation with @dyfer yesterday, regarding backwards compatibility and resulting behavior given installs with mismatched quark and sc3-plugins.

I have a possible solution I'll be testing today that _should_ resolve performance issues if the class and plugin aren't aligned. (I expect a simple check on inputs will do the job.)

If this does work, I'll go ahead and push this change... then I think we'll _really_ be ready to go!

joslloand:
> @telephon I had a conversation with @dyfer yesterday, regarding backwards compatibility and resulting behavior given installs with mismatched quark and sc3-plugins.
> 
> I have a possible solution I'll be testing today that _should_ resolve performance issues if the class and plugin aren't aligned. (I expect a simple check on inputs will do the job.)
> 
> If this does work, I'll go ahead and push this change... then I think we'll _really_ be ready to go!

Ok, here's the fix: https://github.com/supercollider/sc3-plugins/pull/319/commits/8bc45e42374293968a26f0654d2ce4420105ddcf

Which meets these cases, testing cases for matching and unmatching quark AND plugins:

1) matching: **latest test quark** (refactor-foa-nfc-proximity) AND **latest test plugins** (topic/atk-fix-speedOfSound)
--> pass: speed of sound updated to 343 m/s

2) NOT matching: **previous released quark** (v5.0.3) AND **latest test plugins** (topic/atk-fix-speedOfSound)
--> pass: speed of sound remains at 333 m/s [ plugins supply missing speed of sound ]

3) NOT matching: **latest test quark** (refactor-foa-nfc-proximity) AND **previous plugins** (v3.11.1)
--> pass: speed of sound remains at 333 m/s [ plugins cutoff frequency sets equivalent speed of sound, value supplied by class is ignored ]

@joshpar, @dyfer, @telephon, so _now_ I'm thinking we're ready to go with this.

(Finally!)

dyfer:
This seems to work as advertised, thank you!

@joshpar @telephon is it OK to leave that merge commit in?

joshpar:
SHIP IT!

dyfer (approved):
Shippin' it

dyfer:
@joshpar is your change request satisfied? :)

joslloand:
Thanks everyone!

PR #324, opened 2021-09-12
Dfsm shouldn't always post

telephon:
Posting the rules gets particularly problematic when they are very large. You can switch it back on by setting `Dfsm.verbose = true`.


dyfer:
Thanks! Would you be willing to add the helpfile entry for the verbose flag?

telephon:
After another look I've decided that we don't need the post statement, it was only for debugging. I've removed it completely.

telephon:
@dyfer does this solve your request? :)

dyfer (approved):
I does, thanks :)

PR #326, opened 2021-09-30
Add Github action for testing website builds

dyfer:
I've created a github action that runs `bundle install` for PRs and thus should give us heads up in case PRs updating jekyll dependencies break the website.
You can see it [running successfully on my fork](https://github.com/dyfer/sc3-plugins/actions/runs/1292536285), as well as [expectedly failing](https://github.com/dyfer/sc3-plugins/actions/runs/1292543572) when I manually specified an invalid version of one package in `Gemfile.lock`. For these runs I've temporarily made the action to run on all pushes; however here it is set to run on PRs only.

This PR does _not_ provide a way to preview changes made to the website.

### DIFF
--- a/.github/workflows/jekyll.yml
+++ b/.github/workflows/jekyll.yml
@@ -1,0 +1,33 @@
+name: jekyll
+on:
+  # push: # for testing this action
+  pull_request:
+    paths:
+      - '_data/**'
+      - '.bundle/**'
+      - 'assets/**'
+      - 'index.*'
+      - '_config.yml'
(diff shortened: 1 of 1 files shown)


dyfer:
What do you think about this addition @joslloand @gusano ?

joslloand:
A safety check for PRs before pressing go does seem like a good idea to me.

Anyone have any concerns?

dyfer:
Seems there are no concerns... should we merge this?

joshpar (approved):

PR #340, opened 2022-09-24
Reorder JPverbRaw arguments in help file.

tdug:
The class definition and help file argument order did not match. This fixes that, and squashes some warnings that occur when loading the help document.


dyfer (approved):
Sorry for the delay. Thanks!

PR #374, opened 2024-05-22
VBAP with more than 55 channels

lexngu:
Hey there! I'm reaching out to kindly request the merging of my changes to the VBAP UGen to the central repository. This would be helpful for me and my fellow students at the Karlsruhe University of Music as we'd love to use it in rooms equipped with up to 64 loudspeakers. We currently rely on a fork that, in my opinion, needn't exist independently.


telephon (approved):

muellmusik:
Hi @lexngu! Thanks for this. I've actually got a version of VBAP which dynamically allocates memory, and adds some other updates. Just need to clean it up and can do a PR. I think this would be better than just adjusting the hardcoded solution.

muellmusik:
Added at #376. I suggest we merge that instead of this if there are no issues.

lexngu:
Sure, thanks for approval, @telephon, and for your suggestion, @muellmusik ! I also prefer the dynamic memory allocation. I am looking forward to testing the next release!

PR #378, opened 2024-06-15
Cmake force c++ 17

JordanHendersonMusic:
SuperCollider uses c++17 in its code base, not 11 (or earlier), this change forces C++17 compliance. 

https://scsynth.org/t/sc3plugins-keeps-failing-to-compile/9664


capital-G (approved):

JordanHendersonMusic:
@capital-G Do you recon I should merge this myself? The only way I could see this breaking was if someone was using an old compiler that didn't support C++17, but that is a requirement of supercollider so seems very unlikely.

capital-G:
> @capital-G Do you recon I should merge this myself? The only way I could see this breaking was if someone was using an old compiler that didn't support C++17, but that is a requirement of supercollider so seems very unlikely.

I think C++17 gives some interesting new features - as there are official statements that SC target C++17 but it was never *enforced* I think it is fair to do so now, because otherwise we can't use those features.

telephon:
What if someone wanted to compile sc3-plugins for an older version of supercollider?

JordanHendersonMusic:
Didn't think of that... 

I'll add a flag that defaults to 17.

capital-G:
> What if someone wanted to compile sc3-plugins for an older version of supercollider?

They could use an older commit for compiling in this case since SC does officially target C++17 since over 3 years, see https://github.com/supercollider/supercollider/commit/08ecfad7ba6287a7c88e6e82d8122793d48c0c5a 

I think it is currently hard enough to keep the the most recent platforms running (see the broken macOS pipeline and boost upgrade), maintaining older compilers just creates more overhead in the long run for which we don't have enough people to maintain.
Additionally, the versioning of `sc3-plugins` matches the one from `supercollider` most likely not by incident?

> I'll add a flag that defaults to 17.

I'd vote to actually enforce it and push through with it as it may signal that SC are still not fully committed to C++17, which is already a state of 3+ years.

JordanHendersonMusic:
> Additionally, the versioning of sc3-plugins matches the one from supercollider most likely not by incident?

Didn't know that, it's quite a compelling, simple argument --- use the same version across SC and SCplugins.

telephon:
> > What if someone wanted to compile sc3-plugins for an older version of supercollider?
> 
> They could use an older commit for compiling in this case since SC does officially target C++17 since over 3 years, see [supercollider/supercollider@08ecfad](https://github.com/supercollider/supercollider/commit/08ecfad7ba6287a7c88e6e82d8122793d48c0c5a)

Ok, this makes sense to me. Could you please add this information to the README?

JordanHendersonMusic:
Updated readme

dyfer:
Are we fully in c++17 in SC yet? https://github.com/supercollider/supercollider/pull/6246 (I've only skimmed through the issue, sorry if I'm missing something obvious)

JordanHendersonMusic:
@dyfer boost isn't listed in the external libraries folder so that issue should be independent of this.

dyfer:
Ah got it, thanks.

PR #379, opened 2024-08-02
GHA: remove ._ files from macOS zip

metayan:
To get rid of these errors when starting scsynth:

*** ERROR: dlopen '[...]/Extensions/sc3-plugins/SC3plugins/AYUGens/._AY_UGen.scx'
err 'dlopen([...]/Extensions/sc3-plugins/SC3plugins/AYUGens/._AY_UGen.scx, 2): no suitable image found.
Did find: [...]/Extensions/sc3-plugins/SC3plugins/AYUGens/._AY_UGen.scx: file too short

PR #380, opened 2024-08-17
Bump to macos 14 in CI

capital-G:
As mentioned in #379 

Interestingly enough, the macOS artifact zip does not have `._` files in it?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -41,9 +41,9 @@ jobs:
             cache-path: '~/.ccache'
             
           - name: macOS
-            os: macos-11
+            os: macos-14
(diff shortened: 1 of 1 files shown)


metayan:
> As mentioned in #379
> 
> Interestingly enough, the macOS artifact zip does not have `._` files in it?

Yes, this new zip file is fine.
Works here to just unzip and put in `Extensions/` now.
Tested with [pragpub_article.clj](https://github.com/metayan/overtone/blob/tutfix/src/overtone/examples/getting_started/pragpub_article.clj).

Seems like GitHub have fixed something in newer macOS runners.
Thanks for the update.
Closing #379.

dyfer (approved):
Thanks!

PR #381, opened 2024-08-26
GHA: update runners to newer versions

dyfer:
Inspired by the recent macOS runner update, I thought it would be nice to update other runners to newer versions. 
These changes might be preceding the changes to SC's CI, but I'm pretty sure that by the time we are ready for the 3.14 release, we'll be using at least these versions of the deployment tools.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -35,8 +35,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: Linux-x64
+          - name: Linux-x64-22.04
             os: ubuntu-22.04
+            cmake-flags: '-G "Unix Makefiles" -D RULE_LAUNCH_COMPILE=ccache'
+            cache-path: '~/.ccache'
+
+          - name: Linux-x64-24.04
+            os: ubuntu-24.04
             cmake-flags: '-G "Unix Makefiles" -D RULE_LAUNCH_COMPILE=ccache'
             cache-path: '~/.ccache'
             
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -36,27 +36,27 @@ jobs:
       matrix:
         include:
           - name: Linux-x64
-            os: ubuntu-20.04
+            os: ubuntu-22.04
             cmake-flags: '-G "Unix Makefiles" -D RULE_LAUNCH_COMPILE=ccache'
             cache-path: '~/.ccache'
             
           - name: macOS
             os: macos-14
             cmake-flags: '-G Xcode -D CMAKE_OSX_ARCHITECTURES="x86_64;arm64"'
-            xcode-version: '15.3'
-            deployment-target: '10.10'
+            xcode-version: '15.4'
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -36,27 +36,27 @@ jobs:
       matrix:
         include:
           - name: Linux-x64
-            os: ubuntu-20.04
+            os: ubuntu-22.04
```
capital-G, line 39: why not directly move to 24.04? I think it is fair to always use the most recent version of an OS.

dyfer: My hunch here is that _if_ someone were to use the binaries from the GHA build on Linux, a build on an older system is more likely to be compatible with both older and newer OS. But we're not really depending on many OS libraries (FFTW) so this might not be as important. 

Again, this is just a hunch. I'm fine changing it to the newest OS if you don't think there are practical compatibility implications.

capital-G: I think we should currently focus on supporting the latest possible OS on all platforms, so that we are aware of any problems as soon as they occur (maybe even switching to `ubuntu-latest-lts`?).
If the build fails on an older OS, I think it would be valid to just point people to an older SC (plugins) version - and I would somehow doubt that they would miss anything. Maybe the (current) focus on new SC releases is to support current and new platforms, not older ones? See for example the current problems regarding the macOS build and the boost upgrade vs. the somehow abandoned Mingw build issue.
Also, if we always aim for the latest OS, we also have the longest possible period of time where we know that the repo is in a usable state without upgrades, e.g. for Ubuntu 24.04 this would currently be 5 years.

Development efforts are (currently) very sparse, and this balancing act of maintaining a wide variety of new/old OSes/compilers just puts a lot of strain on devs and delays on dev workflows.

dyfer: These are all valid points @capital-G
As I previously indicated, my rationale for using an older builds was so that the build artifacts are runnable on a wider range of OSes. However, we can assume that very few people use these, so that's probably not a major concern here. I'm happy to change this to 24.04 - at the very least, if anybody notices, we'll know if these binaries were used on older Linux systems. 

As for using "latest" images - I think we could consider this, but maybe in a separate discussion, and both for SC and for the plugins. I started writing below, but I think that's better in a new discussion thread: https://github.com/supercollider/supercollider/issues/6430

I wouldn't mind changing this to the `-latest` image, but I'd prefer if we did that consistently with the main SC repo, i.e. discussed it there first and agreed whether we move with that. What do you think?

capital-G: So in the end maybe a matrix with 22.04 and 24.04?

dyfer: I was thinking along the same lines!

dyfer:
I noticed that ccache was not saving cache so I tried to fix it, but it didn't seem to work so I'm reverting to just runner versino update for now.

capital-G (approved):
Thanks for the upgrade! I think we can merge it like this, please do so :)

PR #401, opened 2025-09-19
Add workflow_dispatch trigger to CI workflow

dyfer:
This should allow us to manually trigger Ci for old PRs... maybe?
@capital-G what do you think?


dyfer:
#402 will fix macOS failures

PR #402, opened 2025-09-19
Update CMake minimum version

dyfer:
It seems we need that now... esp on macOS.


dyfer:
Great point about 3.12 in SC! (I chose 3.10 b/c I remembered this was the cutoff to not get errors/warnings in cmake)

PR #405, opened 2025-09-20
build: fix/add ccache support

dyfer:
This PR copies ccache implementation from Supercollider, adding logic to CMakeLists.txt and fixing/adding entries in the CI.

Previously ccache was partially enabled for Linux only, but it didn't work lately, probably due to updated ccache location and possibly updated cmake logic.

Tested here:  https://github.com/dyfer/sc3-plugins/actions/runs/17883065855 (total build time: 2'50"; without ccache it's about 8')

With this repo being rarely updated, the cache is prone to expire, but in the rare instances where there are multiple pushes in shorter succession, tgus improves build times dramatically.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,19 +38,20 @@ jobs:
         include:
           - name: Linux-x64-22.04
             os: ubuntu-22.04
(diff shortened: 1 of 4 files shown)


capital-G (approved):
Thanks :) Seems you are becoming not just a Qt guru but also a CMake guru :)

PR #376, opened 2024-06-10
Update channel limits and VBAP 1.0.3.2

muellmusik:
Adjustments ported from new PD code for VBAP 1.0.3.2. LS amounts allocated dynamically.

PR #435, opened 2026-06-04
VBAP: fix 2D loudspeaker-set parse stride (heap underflow / crash on free)

oudeis01:
### Summary

For 2D speaker setups, `VBAP_Ctor` parses the loudspeaker-set buffer with the wrong per-set stride. `VBAPSpeakerArray.getSetsAndMatrices` (`choose_ls_tuplets` in `vbap.sc`) writes **10** floats per 2D set (2 speaker indices + 4 inverse-matrix + 4 forward-matrix), but the Ctor reads **6** per set and computes `counter = (numvals-2)/((dim*dim)+dim)`. It therefore miscounts the sets and, from the second set on, reads matrix floats as speaker indices. The bad index can be 0, so `final_gs[ls[i]-1]` writes `final_gs[-1]` a heap-buffer underflow. In the
realtime `RTAlloc` pool this silently corrupts an adjacent block, surfacing as a SIGSEGV when the VBAP unit is freed (`VBAP_Dtor` -> `RTFree`).

### Reproduction (SC 3.14.1, sc3-plugins 3.14.0)

```supercollider
s.options.numOutputBusChannels = 12;
s.waitForBoot {
    var azis = [30, -30, 0, 100, -100, 135, -135];   // 7-speaker 2D ring
    var buf  = Buffer.loadCollection(s, VBAPSpeakerArray.new(2, azis).getSetsAndMatrices);
    s.sync;
    SynthDef(\v, {
        var sig = PinkNoise.ar(0.2) * EnvGen.kr(Env.perc(0.01, 0.3), doneAction: 2);
        Out.ar(0, VBAP.ar(7, sig, buf.bufnum, 0, 0, 15));
    }).add;
    s.sync;
    Synth(\v);   // scsynth crashes shortly after the synth frees
};
```

### Root cause (AddressSanitizer)

Building VBAP with `RTAlloc/RTFree` redirected to `malloc/free` (so each allocation gets ASan redzones; the stock SC pool is one big malloc that hides intra-pool overflow) pinpoints it:

    WRITE of size 4 in VBAP_Ctor  VBAP.cpp:713   (final_gs[ls[i]-1] = g[i])
    ... 4 bytes before a 28-byte region          (index -1; final_gs holds numOutputs=7 floats)
    allocated in VBAP_Ctor  VBAP.cpp:637         (final_gs = RTAlloc(numOutputs*sizeof(float)))

Instrumented print right after `vbap(g,ls,unit)`:

    dim=2  numvals=72  x_lsset_amount=11  ->  ls0=0 ls1=0   numOutputs=7

`numvals = 72 = 7*10 + 2` confirms 10 floats per 2D set, but the Ctor computes
`counter = (72-2)/6 = 11` (should be 7).

### Fix

Parse 2D the same way as 3D (stride `(dim*dim*2)+dim`, and read the forward matrix for 2D too). After the fix: `x_lsset_amount = 7`, `ls0=2 ls1=3` (valid, in-bounds), ASan reports zero errors, and the reproduction above plus a churn test (60 spawn/free, spread 0/15/45) run clean.

This also fixes a second latent issue: the `gains_modified` branch uses `x_set_matx[winner_set]` for 2D as well, but 2D never populated `x_set_matx`, so that path previously read uninitialized memory.

### Note

The `// needs to check that numOutputs and x_ls_amount match!!` TODO at `VBAP.cpp:599` is a separate latent bounds issue (when `VBAP.ar`'s numChans < speaker count) and is **not** the cause of this crash (the repro uses 7 == 7); a defensive clamp there would be a reasonable follow-up, though out of scope here.

## Additional changes
@mtmccrea 02.07.26

### .cpp / .sc
- reorder matrix in 2D case for uniformity with 3D
- give setmtx appropriate dimension and indexing
- fix variable initializations and rewrite gain setters for clarity
- rename some vars for clarity
- ctor: Replace unnecessary member initializations with call to VBAP_calc_gain_factors
- Fix memory leak - free m_chanamp in dtor
- Add check that numOutputs == num_ls
- Exit cleanly on data buffer corruption
- Reorg for proper read/write order and legibility (previously a write happened before read)
- Catch loudspeaker data buffer stride error and add more explanatory error message
- Add comments describing loudspeaker buffer contents on both sc and cpp sides

### Help docs
- Add warnings that incomplete loudpseaker coverage is not supported
- remove non-existent method
- add example showing loudspeaker panning curves
- don't set s to internal server (outdated))
- format code examples

### DIFF
--- a/source/VBAPUGens/VBAP.cpp
+++ b/source/VBAPUGens/VBAP.cpp
@@ -646,7 +646,7 @@ static void VBAP_Ctor(VBAP* unit)
 	if(unit->x_dimension == 3)
 		counter = (numvals - 2) / ((unit->x_dimension * unit->x_dimension*2) + unit->x_dimension);
 	if(unit->x_dimension == 2)
-		counter = (numvals - 2) / ((unit->x_dimension * unit->x_dimension) + unit->x_dimension);
+		counter = (numvals - 2) / ((unit->x_dimension * unit->x_dimension*2) + unit->x_dimension);
 	unit->x_lsset_amount=counter;
 
 	if(counter<=0){
@@ -674,7 +674,7 @@ static void VBAP_Ctor(VBAP* unit)
 			unit->x_set_inv_matx[setpointer][i]=buf->data[datapointer++];
 				/*				post("%d",deb++); */
 		}
-		if(unit->x_dimension == 3){
+		if(unit->x_dimension == 2 || unit->x_dimension == 3){
 			for(i=0; i < unit->x_dimension*unit->x_dimension; i++){
 				unit->x_set_matx[setpointer][i]=buf->data[datapointer++];
 
--- a/source/VBAPUGens/VBAP.cpp
+++ b/source/VBAPUGens/VBAP.cpp
@@ -95,17 +95,17 @@ static float atorad = (2.0f * pi) / 360.0f;
 
 struct VBAP : Unit
 {
-	float x_azi;				/* panning direction azimuth */
-	float x_ele;				/* panning direction elevation */
-    float* *x_set_inv_matx;  /* inverse matrice for each loudspeaker set */
-    float* *x_set_matx;      /* matrice for each loudspeaker set */
-    int* *x_lsset;          /* channel numbers of loudspeakers in each LS set */
-	int x_lsset_available;			/* have loudspeaker sets been defined with define_loudspeakers */
-	int x_lsset_amount;			/* amount of loudspeaker sets */
-	int x_ls_amount;					  /* amount of loudspeakers */
-	int x_dimension;					  /* 2 or 3 */
-	float x_spread;							/* speading amount of virtual source (0-100) */
-	float x_spread_base[3];				   /* used to create uniform spreading */
+	float x_azi;                /* panning direction azimuth */
+	float x_ele;                /* panning direction elevation */
+    float* *x_set_inv_matx;     /* inverse matrice for each loudspeaker set */
+    float* *x_set_matx;         /* matrice for each loudspeaker set */
+    int* *x_lsset;              /* channel numbers of loudspeakers in each LS set */
+	int x_lsset_available;      /* have loudspeaker sets been defined with define_loudspeakers */
+	int x_num_lssets;           /* number of loudspeaker sets (pairs or triplets) */
+	int x_num_ls;               /* number of loudspeakers */
+	int x_dimension;            /* 2 or 3 */
+	float x_spread;             /* speading amount of virtual source (0-100) */
+	float x_spread_base[3];     /* used to create uniform spreading */
 	float *final_gs;
 
 	float *m_chanamp;
@@ -243,7 +243,7 @@ static void additive_vbap(float *final_gs, float cartdir[3], VBAP *x)
 // multiple direction panning (source spreading) */
 {
 	float power;
-	int i,j,k, gains_modified;
+	int i,j,k;
 	float small_g;
 	float big_sm_g, gtmp[3];
 	int winner_set;
@@ -252,17 +252,17 @@ static void additive_vbap(float *final_gs, float cartdir[3], VBAP *x)
 	float g[3] = { 0, 0, 0 };
 	int ls[3] = { 0, 0, 0 };
 
-	big_sm_g = -100000.0;
-	best_neg_g_am=3;
+	big_sm_g = -100000.0; /* initial value for largest minimum gain value */
+	best_neg_g_am = dim; /* how many negative values in this set */
 
 	/* BUG: there is a bug that sometimes causes x->x_lsset_amount to be a massive
 		* number.  I haven't tracked it down yet, but its probably an init
 		* bug. 2006-08-13 <hans@at.or.at>
 		*/
 	//	post("x_lsset_amount: %li", x->x_lsset_amount);
-	for(i=0;i<x->x_lsset_amount;i++){
+	for(i=0;i<x->x_num_lssets;i++){
 		small_g = 10000000.0;
-		neg_g_am = 3;
+		neg_g_am = dim;
 		for(j=0;j<dim;j++){
 			gtmp[j]=0.0;
 			for(k=0;k<dim;k++)
@@ -272,30 +272,27 @@ static void additive_vbap(float *final_gs, float cartdir[3], VBAP *x)
 			if(gtmp[j]>= -0.01)
 				neg_g_am--;
 		}
-		if(small_g > big_sm_g && neg_g_am <= best_neg_g_am){
-			big_sm_g = small_g;
-			best_neg_g_am = neg_g_am;
-			winner_set=i;
-			g[0]=gtmp[0]; g[1]=gtmp[1];
-			ls[0]= x->x_lsset[i][0]; ls[1]= x->x_lsset[i][1];
-			if(dim==3){
-				g[2]=gtmp[2];
-				ls[2]= x->x_lsset[i][2];
-			} else {
-				g[2]=0.0;
-				ls[2]=0;
-			}
-		}
+
+        if(small_g > big_sm_g && neg_g_am <= best_neg_g_am){
+            big_sm_g = small_g;
+            best_neg_g_am = neg_g_am;
+            winner_set = i;
+            for(j = 0; j < dim; j++) {
+                g[j] = gtmp[j];
+                ls[j] = x->x_lsset[i][j];
+            }
+            if (dim < 3) { g[2] = 0.0; ls[2] = 0; } // third slot unused for 2D
+        }
 	}
 
-	gains_modified=0;
+    bool src_out_of_bounds = false;
 	for(i=0;i<dim;i++)
 		if(g[i]<-0.01){
-			gains_modified=1;
+            src_out_of_bounds = true;
 		}
 		
     // new from PD
-    if(gains_modified != 1){
+    if(!src_out_of_bounds){
   		power=sqrt(g[0]*g[0] + g[1]*g[1] + g[2]*g[2]);
   		g[0] /= power;
   		g[1] /= power;
@@ -369,16 +366,16 @@ static void spread_it(VBAP *x, float *final_gs)
 		return;
 
 	if(x->x_spread > 70)
-		for(i=0;i<x->x_ls_amount;i++){
+		for(i=0;i<x->x_num_ls;i++){
 			final_gs[i] += (x->x_spread - 70) / 30.0 * (x->x_spread - 70) / 30.0 * 10.0;
 		}
 
-	for(i=0,power=0.0;i<x->x_ls_amount;i++){
+	for(i=0,power=0.0;i<x->x_num_ls;i++){
 		power += final_gs[i] * final_gs[i];
 	}
 
     power = sqrt(power);
-	for(i=0;i<x->x_ls_amount;i++){
+	for(i=0;i<x->x_num_ls;i++){
 		final_gs[i] /= power;
 	}
 }
@@ -389,7 +386,7 @@ static void vbap(float g[3], int ls[3], VBAP *x)
 {
 	/* calculates gain factors using loudspeaker setup and given direction */
 	float power;
-	int i,j,k, gains_modified;
+	int i,j,k;
 	float small_g;
 	float big_sm_g, gtmp[3];
 	int winner_set=0;
@@ -419,20 +416,20 @@ static void vbap(float g[3], int ls[3], VBAP *x)
             x->x_ele += 360.0;
     } else
         x->x_ele = 0.0;
-	
+
+    angle_to_cart(x->x_azi,x->x_ele,cartdir);
+    
 	/* go through all defined loudspeaker sets and find the set which
 		// has all positive values. If such is not found, set with largest
 		// minimum value is chosen. If at least one of gain factors of one LS set is negative
 		// it means that the virtual source does not lie in that LS set. */
 
-	angle_to_cart(x->x_azi,x->x_ele,cartdir);
-	big_sm_g = -100000.0;	/* initial value for largest minimum gain value */
-	best_neg_g_am=3;		  /* how many negative values in this set */
+	big_sm_g = -100000.0;   /* initial value for largest minimum gain value */
+	best_neg_g_am = dim;    /* how many negative values in this set */
 
-
-	for(i=0;i<x->x_lsset_amount;i++){
+	for(i=0;i<x->x_num_lssets;i++){
 		small_g = 10000000.0;
-		neg_g_am = 3;
+		neg_g_am = dim;
 		for(j=0;j<dim;j++){
 			gtmp[j]=0.0;
 			for(k=0;k<dim;k++)
@@ -446,15 +443,11 @@ static void vbap(float g[3], int ls[3], VBAP *x)
 			big_sm_g = small_g;
 			best_neg_g_am = neg_g_am;
 			winner_set=i;
-			g[0]=gtmp[0]; g[1]=gtmp[1];
-			ls[0]= x->x_lsset[i][0]; ls[1]= x->x_lsset[i][1];
-			if(dim==3){
-				g[2]=gtmp[2];
-				ls[2]= x->x_lsset[i][2];
-			} else {
-				g[2]=0.0;
-				ls[2]=0;
-			}
+            for(j = 0; j < dim; j++) {
+                g[j] = gtmp[j];
+                ls[j] = x->x_lsset[i][j];
+            }
+            if (dim < 3) { g[2] = 0.0; ls[2] = 0; } // third slot unused for 2D
 		}
 	}
 
@@ -463,34 +456,35 @@ static void vbap(float g[3], int ls[3], VBAP *x)
 		// gain values. This happens when the virtual source is outside of
 		// all loudspeaker sets. */
 	
-    gains_modified=0;
-    for(i=0;i<dim;i++)
-        if(g[i]<-0.01){
+    bool src_out_of_bounds = false;
+    for(i=0;i<dim;i++) {
+        if(g[i]<-0.01) {
             g[i]=0.0001;
-            gains_modified=1;
+            src_out_of_bounds = true;
         }
+    }
     
-        if(gains_modified==1){
-            new_cartdir[0] =  x->x_set_matx[winner_set][0] * g[0] 
-            + x->x_set_matx[winner_set][1] * g[1]
-            + x->x_set_matx[winner_set][2] * g[2];
-            new_cartdir[1] =  x->x_set_matx[winner_set][3] * g[0] 
-                + x->x_set_matx[winner_set][4] * g[1] 
-                + x->x_set_matx[winner_set][5] * g[2];
-            if (dim == 3) {
-                new_cartdir[2] =  x->x_set_matx[winner_set][6] * g[0] 
-                    + x->x_set_matx[winner_set][7] * g[1]
-                    + x->x_set_matx[winner_set][8] * g[2];
-            } else new_cartdir[2] = 0;
-            cart_to_angle(new_cartdir,new_angle_dir);
-            x->x_azi = (new_angle_dir[0]);
-            x->x_ele = (new_angle_dir[1]);
+    if(src_out_of_bounds) {
+        float * setmtx = &x->x_set_matx[winner_set][0];
+        if (dim == 3) {
+            new_cartdir[0] = setmtx[0] * g[0] + setmtx[1] * g[1] + setmtx[2] * g[2];
+            new_cartdir[1] = setmtx[3] * g[0] + setmtx[4] * g[1] + setmtx[5] * g[2];
+            new_cartdir[2] = setmtx[6] * g[0] + setmtx[7] * g[1] + setmtx[8] * g[2];
+        } else {
+            new_cartdir[0] = setmtx[0] * g[0] + setmtx[1] * g[1];
+            new_cartdir[1] = setmtx[2] * g[0] + setmtx[3] * g[1];
+            new_cartdir[2] = 0.f;
         }
-
-		power=sqrt(g[0]*g[0] + g[1]*g[1] + g[2]*g[2]);
-		g[0] /= power;
-		g[1] /= power;
-		g[2] /= power;
+        
+        cart_to_angle(new_cartdir, new_angle_dir);
+        x->x_azi = (new_angle_dir[0]);
+        x->x_ele = (new_angle_dir[1]);
+    }
+    
+    power = sqrt(g[0]*g[0] + g[1]*g[1] + g[2]*g[2]);
+    g[0] /= power;
+    g[1] /= power;
+    g[2] /= power;
 }
 
 
@@ -508,7 +502,7 @@ static inline void VBAP_calc_gain_factors(VBAP * unit)
 	// only recalculate gain factors if inputs have changed
 	if((azimuth != unit->x_azi) || (elevation != unit->x_ele) || (spread != unit->x_spread)){
 
-		float g[3];
+		float g[3]; // note only 2 slots used in 2D case
 		int ls[3];
 		int i;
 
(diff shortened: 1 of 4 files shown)


dyfer:
Hi @muellmusik, would you be able to review this by any chance?

muellmusik:
> Hi @muellmusik, would you be able to review this by any chance?

Sure. Might take me a bit.

muellmusik:
@oudeis01 Have you done a real world test to confirm that everything sounds correct?

oudeis01:
Hi @muellmusik thanks for asking. and yes, here is a minimal repro plus measured gains.

Same SC code, two builds. Boot `scsynth` against the stock `VBAP.so` vs the patched one , then run:

```
s.options.numOutputBusChannels = 12;
s.waitForBoot {
     var azis = [30, -30, 0, 100, -100, 135, -135];   // 7-speaker 2D ring
     var buf  = Buffer.loadCollection(s, VBAPSpeakerArray.new(2, azis).getSetsAndMatrices);
     s.sync;
     SynthDef(\v, {
         var sig = PinkNoise.ar(0.2) * EnvGen.kr(Env.perc(0.01, 0.3), doneAction: 2);
         Out.ar(0, VBAP.ar(7, sig, buf.bufnum, 0, 0, 15));
     }).add;
     s.sync;
     Synth(\v);
};
```

Tested:
- stock: scsynth crashes shortly after the synth frees (the final_gs[-1] corruption hits RTFree in VBAP_Dtor).
- patched: runs clean, repeatedly.

Gains seem wrong even before the crash. Holding a constant input at fixed azimuths and reading the per-channel output (channel value == VBAP gain), spread = 0:

| source azimuth  | patched                | stock                      |
|-----------------|------------------------|----------------------------|
| 30° (= spk0)    | ch0 = 1.00             | all 0 (silent)             |
| 0° (= spk2)     | ch2 = 1.00             | all 0 (silent)             |
| -30° (= spk1)   | ch1 = 1.00             | ch0 = 0.65 (wrong channel) |
| 15° (spk2↔spk0) | ch0 = 0.71, ch2 = 0.71 | all 0 (silent)             |
| 60° (spk0↔spk3) | ch0 = 0.79, ch3 = 0.61 | all 0 (silent)             |
| 100° (= spk3)   | ch3 = 1.00             | all 0 (silent)             |


(speaker map: ch0=30°, ch1=-30°, ch2=0°, ch3=100°, ch4=-100°, ch5=135°, ch6=-135°)

Across a full -180°..180° sweep (spreads 0/15/45): patched holds constant power Σg² = 1.0 everywhere and matches the analytic VBAP reference (re-derived from the set inverse matrices) to ~1e-7; stock has wide silent dead zones (Σg² → 0) and routes energy to wrong/out-of-range channels. Side-by-side per-channel gain curves (patched left, stock right; dotted = analytic reference):

pr_gain_comparison

-
-
-
One disclaimer: I discovered the issue first in a studio setup with 12ch rig, but my current troubleshooting setup is a 2ch home rig, so I have not yet listened to all 7 channels. I will do the full multichannel listening this week and report back.

oudeis01:
Follow-up to my note:

TL;DR
on the patched build the 2D VBAP sounds correct across all 7 channels. No audible gap or anomaly while sweeping or on hold.
I can also capture per-channel gain plots on this rig if that's useful.

---

I've now run the 7-channel listening test on the studio rig, and the patched build sounds correct across all channels.

**Setup**
- Audient ORIA (USB, 24ch out, 48 kHz, PipeWire-JACK), 7 ear-level loud speakers arranged as the exact 2D ring from the repro: azimuths `[30, -30, 0, 100, -100, 135, -135]` (= L, R, C, Ls, Rs, Lrs, Rrs). The room is 7.0.4; the 4 height speakers are left out here since this PR only touches the 2D path.
- A/B method: same SC session, only `VBAP.so` swapped, via `-U` / `ugenPluginsPath` pointing at a stock-only vs a patched-only plugin dir, so exactly one VBAP loads.
- Heads-up for anyone else A/B-ing plugins: you must reboot the server between swaps, the plugin path is only read at boot.

**Patched (this PR)**
- Smooth, continuous motion all the way around the ring on a slow azimuth sweep; no gaps.
- Each fixed source azimuth localizes to the correct single speaker; a source placed between two speakers images correctly between them.
- `spread` widens the source as expected.
- Everything behaves exactly as the test code intends.

**Stock (current release), same session, VBAP.so swapped back**
- Most fixed azimuths are either silent or come out of the wrong speaker(in my case, the -30 came out from +30 etc.), matching the measured dead-zone / misrouting table above.
- On a moving sweep, scsynth crashes a few seconds in (abrupt cutout + a hard DAC pop). So the crash reproduces by ear.

source/VBAPUGens/VBAP.cpp
```diff
@@ -646,7 +646,7 @@ static void VBAP_Ctor(VBAP* unit)
 	if(unit->x_dimension == 3)
 		counter = (numvals - 2) / ((unit->x_dimension * unit->x_dimension*2) + unit->x_dimension);
 	if(unit->x_dimension == 2)
-		counter = (numvals - 2) / ((unit->x_dimension * unit->x_dimension) + unit->x_dimension);
+		counter = (numvals - 2) / ((unit->x_dimension * unit->x_dimension*2) + unit->x_dimension);
```
mtmccrea: If the counter math is the same regardless of the dimension, the `if`s aren't needed. Same with the change below. 
However, making the calculations identical regardless of the dimension seems odd... why were they different in the first place? @muellmusik

mtmccrea:
Here's a straightforward way to plot speaker gains using the minimal repro settings you provided above:
```supercollider
(
s.waitForBoot {
	~azis = [30, -30, 0, 100, -100, 135, -135];   // 7-speaker 2D ring
	~buf  = Buffer.loadCollection(s, VBAPSpeakerArray.new(2, ~azis).getSetsAndMatrices);
}
)

(
{
	VBAP.ar(7, 
		DC.ar(1), // signal
		~buf.bufnum, // VBAP data
		Line.kr(-180,180,1), // position
		0, 0);
}.plot(1).superpose_(false);
)
```

For me (macOS, SC 3.14), the provided example does not break/crash:

Screenshot 2026-06-07 at 12 19 23

Aside: In case it's relevant, what are data are you plotting in your above plots? Is it from SC output? looks like matplotlib/python?

source/VBAPUGens/VBAP.cpp
oudeis01: Agreed on the cleanup: 

after this fix the 2D and 3D counter expressions are identical

`(numvals-2)/(dim*dim*2 + dim)`

and the `if(dim==2 || dim==3)` below is effectively unconditional since `dim` is only ever 2 or 3, so both can collapse to a single branch. I kept them split here only to mirror the existing structure and keep the diff minimal; happy to squash them, but thinking it would be better after I clarify the bug sufficiently.

On *"why were they different in the first place"* :

as far as I can trace, they were not, until #376 ("Update channel limits and VBAP 1.0.3.2", 2025-10-21).
That commit added the forward (non-inverted) matrix to the 2D layout on the class side

`vbap.sc`: `list_length = amount * 6 + 2` -> `amount * 10 + 2`, plus the new `mat[]` write and its computation in `calc_2D_inv_tmatrix`

and on the C side it updated the 3D reader to the wider stride and added the 3D forward-matrix read, but the 2D reader was left at the old `/6` stride with the forward-matrix read still gated by `if(dim==3)`. So the 2D/3D asymmetry looks like the 2D read path simply not being updated alongside the 2D write path in #376, rather than an intentional difference.

I agree taht it would be much clearer if @muellmusik would confirm the intended 2D layout and whether the 2D reader was meant to match the 3D one here. If so, collapsing the branches as @mtmccrea suggests is the natural finish.

oudeis01:
Hello @mtmccrea, thank you for the code and the plot. 

I would like to break down my answer to two:

On my figures:

you are right that I used python for the plotting, though the capture is done in SC, not python. I ran a headless sclang that feeds `DC.ar(1)` through `VBAP.ar`, so with a constant unit input each output channel settles to exactly that channel's VBAP gain. Then I read the 7 gains back with `A2K.kr` + `SendReplay.kr` over osc and sweep azimuth x spread to a csv file. Python(matplotlib) is used only for plotting the CSV and the VBAP reference(dotted) lines. So as far as I see, it is the same mechanism as your `.plot` which records the running UGEn internally on the server.

---

On why it does not break for you: 

I think it is a release skew, not a platform dependent luck/unluck. The 2D buffer layout changed in #376. 

`vbap.sc` went from `amount * 6 + 2` to `amount * 10 + 2` (it now also writes the forward matrix per 2D set), but the C 2D reader still uses the `/6` stride and only reads the forward matrix under `if(dim==3)`.

With the 10-float layout read at a 6-float stride, the first set parses fine and every later set reads matrix floats as speaker indices, which is what produces the dead zones here and, on this Linux box, a heap-buffer-overflow write into the gains array (ASAN flags it at VBAP_Ctor, gains[-1]).

#376 first shipped in sc3-plugins 3.14.0 (2026-05-17), the previous release 3.13.0 predates it. So a pre-3.14.0 install has `vbap.sc` at 6/set and the C reader at 6/set, which agree, and everything sounds correct.

I note that you mention your SC version but not the sc3-plugins version (my box is sclang 3.14.1 with sc3-plugins 3.14.0).
Could you check your sc3-plugins version, and/or run this language-only probe (no server, cannot crash) and report `raw.size`?

```supercollider
(
var azis = [30, -30, 0, 100, -100, 135, -135];
var raw = VBAPSpeakerArray.new(2, azis).getSetsAndMatrices;
("raw.size = " ++ raw.size ++ "   sets/10 = " ++ ((raw.size - 2) / 10)).postln;
)
```

I prediction is that, if you get a `raw.size = 44` then your `vbap.sc` is the pre-#376 6/set layout (consistent with the old C reader, therefore clean).
If you get `72`(which is my case) you are on the new 10/set layout, in which case your clean `.plot` is more surprising and requires a closer look on my side. 

Btw, on 3.14.0 here it crashes with your exact code: the `.plot` synth boots, then frees on doneAction:2 and the server goes down, so the plot window opens empty.

oudeis01:
This will reproduce the dead-zone / crash behaviour of what I have reported in the PR.
The code I used had a little more lines that swaps the VBAP.so and reboots the server, but this one below will load the VBAP in your installed sc3-plugin and run once, save a single csv.

```supercollider
(
var here, azis, azGrid, spreads, settle;
var server, buf, synth, csvRows, t0, log, stride, nSets, raw;
var gotReply, lastReply, sumsqLog;

here    = thisProcess.nowExecutingPath.dirname;
azis    = [30, -30, 0, 100, -100, 135, -135];   // canonical 7-speaker 2D ring (the one that triggers the bug)
azGrid  = (-180, -175 .. 180);                   // 73 probe directions
spreads = [0, 15, 45];
settle  = 0.14;                                   // > VBAP slew, < anything slow

t0  = Main.elapsedTime;
log = {|m| ("[" ++ (Main.elapsedTime - t0).round(0.01).asString ++ "] " ++ m).postln };

csvRows  = List["spread,azimuth,g0,g1,g2,g3,g4,g5,g6,sumsq"];
sumsqLog = List[];   // (spread0 only) sum of squares per azimuth, for the verdict

server = Server.default;
server.options.numOutputBusChannels = 16;
server.options.memSize   = 262144;
server.options.numBuffers = 1024;
server.options.blockSize  = 64;

log.("SuperCollider " ++ Main.version);

server.waitForBoot {
    Routine {
        var arr;
        log.("server booted, default plugins (your installed VBAP.so)");

        arr = VBAPSpeakerArray.new(2, azis);
        raw = arr.getSetsAndMatrices;

        // --- self-report the layout (the macOS vs Linux discriminator) ---------
        // pre-#376 sclang writes 6 floats/set (raw.size = amount*6 + 2),
        // #376+ sclang writes 10 floats/set (raw.size = amount*10 + 2).
        stride = if(((raw.size - 2) % 10) == 0) { 10 }
                 { if(((raw.size - 2) % 6) == 0) { 6 } { -1 } };
        nSets  = if(stride > 0) { ((raw.size - 2) / stride).asInteger } { -1 };
        "".postln;
        "=== layout self-report =====================================".postln;
        ("raw.size (numvals)   = " ++ raw.size).postln;
        ("detected stride      = " ++ stride ++ " floats/set"
            ++ if(stride == 10) { "   (#376+ : sc3-plugins >= 3.14.0)" }
               { if(stride == 6) { "   (pre-#376 : sc3-plugins <= 3.13.x)" } { "   (UNRECOGNISED)" } }).postln;
        ("real loudspeaker sets = " ++ nSets).postln;
        "============================================================".postln;
        "".postln;

        buf = Buffer.loadCollection(server, raw);
        server.sync;

        SynthDef(\vbapProbe, {|azimuth = 0, spread = 0, snap = 0|
            var sig = DC.ar(1);
            var pan = VBAP.ar(7, sig, buf.bufnum, azimuth, 0, spread);
            SendReply.kr(Trig1.kr(snap, 0.01), '/gains', A2K.kr(pan));
            // no Out: silent capture
        }).add;
        server.sync;

        gotReply = false; lastReply = nil;
        OSCdef(\g, {|msg| lastReply = msg[3..9]; gotReply = true }, '/gains');

        synth = Synth(\vbapProbe, [\azimuth, 9999, \spread, 0], server);  // odd start so first set sticks
        server.sync; 0.2.wait;

        spreads.do {|sp|
            log.("sweep spread=" ++ sp);
            azGrid.do {|az|
                var g, ssq;
                synth.set(\spread, sp, \azimuth, az);
                settle.wait;
                gotReply = false;
                synth.set(\snap, 1);
                (0..40).do {|tries|
                    if(gotReply.not and: { server.serverRunning }) { 0.01.wait };
                };
                synth.set(\snap, 0);

                if(server.serverRunning.not) {
                    "".postln;
                    ("*** SERVER DIED at spread=" ++ sp ++ " az=" ++ az
                        ++ " -> this is the crash; you are on the broken layout ***").postln;
                    File.use(here +/+ "vbap_gains_repro.csv", "w", {|f|
                        csvRows.do {|r| f.write(r ++ "\n") } });
                    ("wrote partial " ++ (here +/+ "vbap_gains_repro.csv")).postln;
                    // 0.exit;   // uncomment for headless
                    nil.yield;
                };

                g   = if(gotReply) { lastReply } { Array.fill(7, { \nan }) };
                ssq = if(gotReply) { g.collect {|x| x * x }.sum } { \nan };
                if(sp == 0) { sumsqLog.add(ssq) };
                csvRows.add(
                    ([sp, az] ++ g ++ [ssq]).collect {|x|
                        if(x == \nan) { "nan" } { x.asString }
                    }.join(",")
                );
            };
        };

        // --- write CSV next to this file (no out/ dir needed) ------------------
        File.use(here +/+ "vbap_gains_repro.csv", "w", {|f|
            csvRows.do {|r| f.write(r ++ "\n") } });
        log.("wrote " ++ (here +/+ "vbap_gains_repro.csv") ++ "  (" ++ (csvRows.size - 1) ++ " rows)");

        // --- in-language verdict (no Python needed) ----------------------------
        if(sumsqLog.size > 0) {
            var clean = sumsqLog.select {|x| x.isNumber };
            var dead  = clean.count {|x| x < 0.5 };       // constant-power should keep this ~1.0
            "".postln;
            "=== verdict (spread=0, constant-power check) ===============".postln;
            ("sum(g^2) min / max   = " ++ clean.minItem.round(0.001) ++ " / " ++ clean.maxItem.round(0.001)
                ++ "   (ideal = 1.0 everywhere)").postln;
            ("dead-ish directions   = " ++ dead ++ " / " ++ clean.size ++ "   (sum(g^2) < 0.5)").postln;
            if(dead > 0) {
                "=> DEAD ZONES PRESENT: this install mis-parses the 2D sets (the bug).".postln;
            } {
                "=> no dead zones: this install parses the 2D sets correctly.".postln;
            };
            "============================================================".postln;
        };

        synth.free;
        log.("DONE (server left running; synth freed)");
        // 0.exit;   // uncomment for a headless one-shot
    }.play;
};
)
```

mtmccrea:
Thanks for your thorough explanation, @oudeis01.

> I prediction is that, if you get a raw.size = 44 then your vbap.sc is the pre-https://github.com/supercollider/sc3-plugins/pull/376 6/set layout (consistent with the old C reader, therefore clean).

Indeed I get `raw.size = 44`. So I'll be sure not to update my plugins until this gets fixed ;-)

source/VBAPUGens/VBAP.cpp
muellmusik: > However, making the calculations identical regardless of the dimension seems odd... why were they different in the first place? [@muellmusik](https://github.com/muellmusik)

That's a good question! As this was a port from the PD update, I'm really not sure tbh, but I strongly suspect it came from the PD code.

muellmusik:
This looks reasonable to me, but would be good to just test here as well. I think probably just a bug introduced with one of the updates. Apologies for that. I originally had my own release for this, but IIRC an early version was added here with some changes and without checking with me. It was rather later I got back to trying to reconcile this with updates from the PD version.

dyfer:
How do we proceed here?
As this seems to be a regression from plugins v. 3.13, I think it would be good to merge this sooner than later and release 3.14.1 bugfix release of the plugins.

muellmusik:
I’d like to test this but to be honest will be busy until mid-July. Happy for this to be merged before but would be good if there was another review in that case.

dyfer:
Thanks @muellmusik 
@mtmccrea would you be able to review this by any chance?

mtmccrea:
Sure, I'll have a closer look in the next couple of days.

mtmccrea:
I mistakenly pushed some work in progress on this... I'll clean that up and let you know when it's ready to be looked at.

mtmccrea:
Just dropping an update here. After looking into this in detail, there are actually quite a number of issues that came up regarding the handling of the 2D vs. 3D case. It also led to finding broken edge cases in the case of speaker layouts without full coverage.

That could take yet more work to fix, and given that there now seems to be release decisions waiting on this, I'll try to isolate the changes that pertain only to the handling of 2D vs. 3D VBAP (the topic of this PR), and perhaps leave the broken behavior of incomplete layouts to another PR.

dyfer:
Thanks @mtmccrea 

> there are actually quite a number of issues that came up regarding the handling of the 2D vs. 3D case. It also led to finding broken edge cases in the case of speaker layouts without full coverage.

Do you happen to know if these were introduced by https://github.com/supercollider/sc3-plugins/pull/376 or has it always been like that? I also understand if it's not worth the time to investigate an old PR.

mtmccrea:
I'm pretty sure it was always like that. 

It's a very old implementation of VBAP. I'm curious whether a different reference implementation (like Politis's matlab library) would more gracefully handle these edge cases.

Triangulation could also be more reliably done with the Delaunay method, but that may not be readily available in a library we already use like BOOST.

mtmccrea:
OK i've pushed my changes now, which I think should cover the full scope of handling the different loudspeaker config buffer payload. I'll update the description to include these changes.

Not included in these changes is a fix for insufficient loudspeaker coverage. I started on this, but can't know when/if that job will get finished any time soon. A more robust solution involves a presumably more elaborate fix where virtual loudspeakers are added internally to fill in gaps, but are not sent out. An informed user would be able to do this manually on the SC side, so an interim solution is to add such an example to help docs.

mtmccrea:
Now the question of how review should proceed...
Now that I've touched a bunch of code, someone else should probably look over it.

At the very least, @oudeis01 could test and confirm that these changes appropriately build on their PR.

mtmccrea:
fwiw I've tested various valid 2D and 3d configurations, visually inspecting gains with a plotted panning signal, and I don't observe any discontinuities or anomalies.

oudeis01:
Thanks @mtmccrea for the careful follow-up commits, especially the reprojection fix in d26587c0, which my original patch did not cover. I built and tested the PR head (6ebae7c) and can confirm these changes build correctly on top of the original fix.

**Environment**

- Arch Linux, SuperCollider 3.14.1, GCC 15
- Clean cmake build against the 3.14.1 plugin headers, no warnings from VBAP.cpp
- Both the classlib and the plugin were swapped to the PR head in an isolated path, so nothing from the installed 3.14.0 was mixed in

**Gain invariance (2D, in-coverage path)**

I re-ran the sweep from my earlier report: 7-speaker ring `[30, -30, 0, 100, -100, 135, -135]`, azimuths -180..180 in 5 degree steps, spreads 0/15/45, `DC.ar(1)` probe read back via `SendReply`.

- All 219 measurements are bitwise identical to the original two-line fix
- Per-row energy sums stay at 1.0 throughout, no NaN anywhere

This matches what I expected from reading d26587c0: the 2D inverse matrix is numerically unchanged by the reordering, so in-coverage panning is unaffected, and the layout change corrects the out-of-coverage reprojection branch.

**Ctor validation**

The new numOutputs check works as intended. Tested with `VBAP.ar(6, ...)` against a 7-speaker buffer:

- Server prints `VBAP Error: ugen's numOutputs (6) does not match the number of loudspeakers reported in the loudspeaker buffer (7)`
- The unit stays silent, and `.set` plus `.free` on the failed unit are safe (server stays alive)
- One note: this turns a previously silently-broken case into an explicit no-op, so it may be worth a line in the release notes as a user-visible behavior change

**One regression I noticed while reviewing 4f18c4f**

The old ctor initialized

```c
unit->x_spread_base[0] = 0.0;
unit->x_spread_base[1] = 1.0;
unit->x_spread_base[2] = 0.0;
```

and this initialization was dropped in 4f18c4f without being restored. In the 3D path with spread > 0, the first gain calculation now reads uninitialized real-time memory:

- `spread_it()` calls `new_spread_dir()`, which reads `x_spread_base` (dot product, then `acos`) before `new_spread_base()` has ever written it
- If that memory happens to contain garbage or NaN, `gamma` becomes NaN, and the `fabs()` fallback does not trigger (NaN comparisons are all false)
- The resulting spreaddir is NaN, and `new_spread_base()` then stores NaN back into `x_spread_base`, so the gains can stay NaN for the lifetime of the synth

2D is unaffected (the 2D spread path never touches `x_spread_base`), which is why the sweeps above cannot exercise it. Restoring the three-line initialization in the ctor (next to the `x_azi`/`x_ele` NaN init) should resolve it.

**Minor cosmetic note**

`additive_vbap()` still initializes `best_neg_g_am` with 3 while `vbap()` now uses `dim`. Harmless for dim == 2 (at most 2 negative gains, so the first set still wins the comparison), just an inconsistency.

mtmccrea:
Thanks for having a close look @oudeis01!

I've addressed the two issues you raised:
- Restored the initialization of `x_spread_base`
- Updated `additive_vbap()` to mirror `vbap()` in its use of the dimension variable. There's an opportunity to refactor these two functions to avoid repeated code that can lead to this kind of mismatch in the future. But after looking into it a bit, I don't want to touch this much more and depart from the original author's code without having headspace to do it properly.

So I think this is good to go!

I'll reiterate your **release note** suggestion for the merging reviewer:

*We now check that the number of speakers specified matches the data buffer: this turns a previously silently-broken case into an explicit no-op*

dyfer:
Thanks for your work @mtmccrea and @oudeis01 
@muellmusik would you be able to review this by any chance? It seems that this is now functionally sound. Please let me know if I can help with testing etc.

muellmusik:
> @muellmusik would you be able to review this by any chance?

Hi. Sorry I've had a bit of a family crisis going on, and haven't had much space for this. This looks good to me, but I've not had a chance to test. I think it can probably be merged as I don't want to delay it much further. @dyfer have you reviewed it fully?

dyfer:
@muellmusik I have not :) I don't necessarily have the resources to really evaluate cpp changes, I'd prefer for someone else to evaluate them.

Based on the conversation I'm confident in the work of @mtmccrea and @oudeis01. If you feel this needs a proper cpp review, then that would have to be done by someone else. If this only needs a general test and looking over, I can do that.

mtmccrea:
Based on what Scott said, and the review/testing we've done, I think a once-over would suffice to finalize this?

> This looks good to me, but I've not had a chance to test. I think it can probably be merged as I don't want to delay it much further.

muellmusik:
Sorry. Things are not good, so I can't help. I trust your collective judgement.

oudeis01:
@dyfer @mtmccrea  I think this is ready, and what's left is procedural rather than technical.

The cross-review Scott asked for on Jun 17 has happened in both directions: @mtmccrea reviewed and extended my original fix, and I reviewed his commits (catching the dropped `x_spread_base` init, which he's since restored). Both of us have tested independently. His plotted panning inspection across 2D and 3D configs, my rebuilt sweep on the PR head plus a 7-speaker listening test. Scott has said he's happy for it to be merged without his own testing.

@dyfer, you mentioned you could do a general test and look-over. If that's enough alongside the above, I think it can go in. It's a hard crash on 3.14.0 for any 2D VBAP user, so a 3.14.1 bugfix release would be worth it.

One loose end: the suggestion to collapse the now-identical 2D/3D counter branches. Scott's note that the asymmetry likely came from the PD port settles the open question. I'll squash them if you want it here, or leave the structure as-is for a minimal diff. Just say which, I'm glad to do it.

Happy to open a separate issue for the insufficient-coverage case and draft the interim help-doc example.

mtmccrea:
>I'll squash them if you want it here, or leave the structure as-is for a minimal diff. Just say which, I'm glad to do it.

This has already been taken care of with my additional changes (see `numSets` variable calculation).

I've also squashed my commits for cleaner history.

capital-G:
Haven't followed the PR yet, but there are people in the forum who ran into this problem and this provided a fix -> https://scsynth.org/t/vbap-issue/13542/5?u=dscheiba

Any ideas how to get progress on it?

@dyfer do we want to make a patch release once this is merged, so we have a working version of VBAP with the old server API?

mtmccrea (approved):
At the risk of breaking the guideline of not approving a PR I've contributed to, I'll sign off on this one. At least a few people involved here have informally signed off on moving this forward, it's had a good amount of testing, and I stepped through a copilot review for a sanity check.

dyfer:
Thanks for additional reference @capital-G 

> @dyfer do we want to make a patch release once this is merged, so we have a working version of VBAP with the old server API?

Yes, I'd like to do that.
My plan is to do a final review on this (basically re-read the comments from @oudeis01 and @mtmccrea and build/test locally). Then I'll merge it and we can release sc3-plugins 3.14.1 (with an RC or not?)

